PR #83, opened 2015-01-26
Synchronized  UnsignedXYTypes

dietzc:
parallel access on these types caused problems. it was fixed for BitType, but not for the remaining types.

See https://github.com/imglib/imglib2/tree/synchronized-bit-access for details.

### DIFF
--- a/src/main/java/net/imglib2/type/AbstractBit64Type.java
+++ b/src/main/java/net/imglib2/type/AbstractBit64Type.java
@@ -108,20 +108,22 @@ public abstract class AbstractBit64Type<T extends AbstractBit64Type<T>> extends 
 		final long k = i * nBits;
 		final int i1 = (int)(k >>> 6); // k / 64;
 		final long shift = k & 63; // Same as k % 64;
-		final long v = dataAccess.getValue(i1);
-		if (0 == shift) {
-			// Number contained in a single long, ending exactly at the first bit
-			return v & mask;
-		} else {
-			final long antiShift = 64 - shift;
-			if (antiShift < nBits) {
-				// Number split between two adjacent long
-				final long v1 = (v >>> shift) & (mask >>> (nBits - antiShift)); // lower part, stored at the upper end
-				final long v2 = (dataAccess.getValue(i1 + 1) & (mask >>> antiShift)) << antiShift; // upper part, stored at the lower end
-				return v1 | v2;
+		synchronized ( dataAccess ) {
+			final long v = dataAccess.getValue(i1);
+			if (0 == shift) {
+				// Number contained in a single long, ending exactly at the first bit
+				return v & mask;
 			} else {
-				// Number contained inside a single long
-				return (v >>> shift) & mask;
+				final long antiShift = 64 - shift;
+				if (antiShift < nBits) {
+					// Number split between two adjacent long
+					final long v1 = (v >>> shift) & (mask >>> (nBits - antiShift)); // lower part, stored at the upper end
+					final long v2 = (dataAccess.getValue(i1 + 1) & (mask >>> antiShift)) << antiShift; // upper part, stored at the lower end
+					return v1 | v2;
+				} else {
+					// Number contained inside a single long
+					return (v >>> shift) & mask;
+				}
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


tpietzsch:
@dietzc: Unsigned128BitType does not need to be synchronized. There is no conflict there: Each value spans exactly two longs, so no element of the underlying primitive array is shared by more than two values. Please remove this commit.
The remaining type that needs to be synchronized (as far as I can see) is ```AbstractBit64Type``` (which is extended by ```BasePairBitType``` and ```UnsignedVariableBitLengthType```). Could you add synchronization for that as well?

dietzc:
done. build still works. something else missing?

tpietzsch:
@dietzc Hmm, you overdid it a bit... Get()s don't have to be synchronized. I fixed that and also moved some computations out of the synchronize blocks. Can you please verify that your stuff works with this?

tpietzsch:
@dietzc Did you run the KNIP test already?

dietzc:
nope, not yet! I will test it today!

dietzc:
:+1: go!

can you release imglib2 and imglib2-algorithm?

tpietzsch:
@dietzc I would wait until the morphology package is merged before releasing imglib2-algorithm.

acardona:
Hi @tpietzsch , @dietzc , wouldn't it have made more sense to create extended classes to overwrite the methods to be synchronized? E.g. SynchronizedUnsigned4BitType extends Unsigned4BitType ... and then extend just the methods that need synchrized (dataAccess) ? The way you are doing it now forces all uses of these types to be synchronized, even when the synchronization is not necessary.

tpietzsch:
@acardona @dietzc If one of you wants to give this a try please go ahead!

dietzc:
I won't find time to do that. I also don't understand how this would work (even if it would be nice to have it). If I get a `RandomAccessibleInterval<T>` with `T extends RealType<T>` as the input for my algorithm which is paralellized, how would I change the type of the `T` to something which is thread-safe (or not)?

acardona:
@dietzc Picture the case where it is better to process multiple images with one thread each rather than one single image with many threads. For the earlier case the synchronization is an unnecessary overhead.

I agree though that the default approach is to assume types are internally self-consistent when it comes to read/write into their underlying container. So I would favor the approach where an "unsafe" subpackage contains the fast UnsynchronizedXYBitType versions, differing only in the write methods.

@tpietzsch By the way the get method is also unsafe and should be synchronized given that it reads from a container that can be written to simultaneously by another thread in ways that do not guarantee atomicity at the single array slot entry point like for the other types.

dietzc:
@acardona I understand the use-case, this actually our main use-case with KNIME :-). But if I understand correctly, then you have to decide which type you want to use right when the image is created (e.g. in the `ImgOpener`) and this type is then propagated through your complete program (assuming you don't add any instance-of checks and use algorithms which are independent of the type). Right?

tpietzsch:
@acardona That is ok. If one pixel is touched by multiple threads at the same time, the results are undefined. In this respect this is similar to ```ComplexDoubleType```. If you write and read the same pixel of ```ComplexDoubleType``` from two threads, you may also see inconsistent results because writes are not atomic. The difference in the BitTypes is that the write touches multiple pixels. Assume that four bits from the same long are ```0100``` and thread A wants to reads the second bit of those while thread B sets the fourth bit to ```1```. Then it doesn't matter whether thread A sees the value before or after the write. It gets ```0100``` or ```0101``` and it doesn't matter which because in the second bit it is unchanged.
For the imglib2 types we implicitly guarantee the same behaviour as primitive arrays. For example, if you have a ```long[] array``` and simultaneously read and write ```array[0]``` it is not guaranteed that the read does not happen in the middle of the write, because ```long``` writes are not atomic. It is guaranteed however that simultaneously writing ```array[0]``` and ```array[1]``` will not influence each other. As I said, this is an implicit guarantee which is just the interpretation of how all the types behaved before the fraction branch was merged. The fraction branch broke the behaviour and this PR fixed it. I think the implicit guarantee should be made explicit in the documentation.

acardona:
@dietzc For the Unsigned12BitType yes. For the others, my only use has been for special-purpose applications like storing feature vectors in a compact manner, or for avoiding using a full long because an integer range is no enough for e.g. integral histograms. So: more for results of computations rather than for reading files from disk.

acardona:
@tpietzsch OK great, just thought of pointing that out.

dietzc:
@acardona ah! this absolutely makes sense. thanks for clarifying.

tpietzsch:
@dietzc Do you have a small multi-threaded example that I could use to test correctness and performance when evaluating the synchronized vs a lock-free approch?

dietzc:
unfortunately not, I always tested it with our KNIME test-flows.

PR #84, opened 2015-01-26
ImgView: add static wrap method to represent a RAI as an Img

dietzc:
The methods makes sure, that we only wrap a RandomAccessibleInterval if
its not already an Img


tpietzsch:
thanks @dietzc!

PR #88, opened 2015-05-05
added most simple implementation of integer shear transform

hanslovsky:
This transforms an n-dimensional coordinate c like this:
```java
c[ shearDimension ] = c[ shearDimension ] + c[ referenceDimension ]  // forward
c[ shearDimension ] = c[ shearDimension ] - c[ referenceDimension ]  // backward
```
Source and target dimensions are the same.

### DIFF
--- a/src/main/java/net/imglib2/transform/integer/shear/AbstractShearTransform.java
+++ b/src/main/java/net/imglib2/transform/integer/shear/AbstractShearTransform.java
@@ -21,7 +21,8 @@ import net.imglib2.transform.integer.BoundingBoxTransform;
  * @author Philipp Hanslovsky <hanslovskyp@janelia.hhmi.org>
  *
  */
-public abstract class AbstractShearTransform implements InvertibleTransform, BoundingBoxTransform  {
+public abstract class AbstractShearTransform implements InvertibleTransform, BoundingBoxTransform
+{
 	
 	
 	protected final int nDim;
--- a/src/main/java/net/imglib2/transform/integer/shear/AbstractShearTransform.java
+++ b/src/main/java/net/imglib2/transform/integer/shear/AbstractShearTransform.java
@@ -1,0 +1,124 @@
+/**
+ * 
+ */
+package net.imglib2.transform.integer.shear;
+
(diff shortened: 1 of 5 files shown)


axtimwalde:
Thanks @hanslovsky!  Can you add the methods

    RandomAccessibleInterval Views.shear(RandomAccessibleInterval, int, int)
    RandomAccessibleInterval Views.unshear(RandomAccessibleInterval, int, int)
    RandomAccessible Views.shear(RandomAccessible, int, int)
    RandomAccessible Views.unshear(RandomAccessible, int, int)

?  I think, then we're done and ready to merge.

Thanks

hanslovsky:
I will do that.
As the Interval implementations will require Views.extend to avoid out of bounds errors, I will create
```java
RandomAccessibleInterval Views.shear(RandomAccessibleInterval, int, int, T) // use extendValue( source, T t )
RandomAccessibleInterval Views.shear(RandomAccessibleInterval, int, int ) // convenience for RealType, extend with Double.NaN
```

axtimwalde:
@hanslovsky you're right, the dimensions change.  Instead of
    RandomAccessibleInterval Views.shear(RandomAccessibleInterval, int, int)
    RandomAccessibleInterval Views.unshear(RandomAccessibleInterval, int, int)
I suggest that we have
    RandomAccessibleInterval Views.shear(RandomAccessibel, Interval, int, int)
    RandomAccessibleInterval Views.unshear(RandomAccessible, Interval, int, int)
which would be more generic.  No shortcuts for RealType and NaN, RealType is not always Double and may not have NaN (intrinsic problem of RealType write inheritance being the same as read inheritance).

hanslovsky:
I added the convenience methods to Views.

axtimwalde:
@hanslovsky Couldn't find

    RandomAccessible Views.shear(RandomAccessible, int, int)
    RandomAccessible Views.unshear(RandomAccessible, int, int)

hanslovsky:
Added that now.

axtimwalde:
Thanks @hanslovsky this is a very useful addition!

PR #89, opened 2015-05-11
PermutationTransform

hanslovsky:
Added the PermutationTransform, which transforms either a single or all coordinates of a RandomAccessibleInterval according to a bijective permutation over the index set of that coordinate.

Implemented the following classes/methods:

```java
class AbstractPermutationTransform // with test
class PermutationTransform  // with test
class SingleDimensionPermutationTransform // with test
IntervalView Views.permute( RandomAccessibleInterval, int[] )
IntervalView Views.permuteOneDimension( RandomAccessibleInterval, int[], int )
IntervalView Views.permuteInverse( RandomAccessibleInterval, int[] )
IntervalView Views.permuteInverseOneDimension( RandomAccessibleInterval, int[], int )
```

### DIFF
--- a/src/main/java/net/imglib2/transform/integer/shear/AbstractShearTransform.java
+++ b/src/main/java/net/imglib2/transform/integer/shear/AbstractShearTransform.java
@@ -1,0 +1,125 @@
+/**
+ * 
+ */
+package net.imglib2.transform.integer.shear;
+
+import net.imglib2.Localizable;
+import net.imglib2.Positionable;
+import net.imglib2.transform.InvertibleTransform;
+import net.imglib2.transform.integer.BoundingBoxTransform;
+
+/**
+ * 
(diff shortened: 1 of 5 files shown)


src/main/java/net/imglib2/transform/integer/permutation/AbstractPermutationTransform.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+public abstract class AbstractPermutationTransform implements
+		InvertibleTransform {
+	
+	final protected int[] lut;
+	final protected int[] inverseLut;
+	
+
+	/**
+	 * @param lut must be a bijective permutation over its index set, i.e. for a lut of legnth n,
```
axtimwalde: Typo: length

axtimwalde:
Moved to https://github.com/imglib/imglib2/pull/90

PR #92, opened 2015-06-14
Point: add equals method

dietzc:
Add equals method to Point.


tpietzsch:
@dietzc thank you! Could you also add an equals() for RealPoint?

PR #94, opened 2015-07-01
Type bug fixes

awalter17:
Hello! The first commit is to prevent ArrayIndexOutOfBounds exceptions when creating an Unsigned128BitType with less than 7 bytes. And the second commit is to prevent ClassCastException. 

Please let me know if there are any issues with either of these commits!

### DIFF
--- a/src/main/java/net/imglib2/type/numeric/integer/Unsigned128BitType.java
+++ b/src/main/java/net/imglib2/type/numeric/integer/Unsigned128BitType.java
@@ -142,34 +142,22 @@ public class Unsigned128BitType extends AbstractIntegerType<Unsigned128BitType> 
 		bytes[15] = (byte)((lower >>>  8) & 0xffL);
 		bytes[16] = (byte) (lower         & 0xffL);
 	}
-	
+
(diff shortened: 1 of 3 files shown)


tpietzsch:
@awalter17 Thanks for the bug fix!
Do you have test code for the Unsigned128BitType ```set(byte[])``` method?
Your fix looks fine, but I think the whole method could be shortened to
```
	public void set( final byte[] bytes ) {
		final int k = i * 2;
		int b = bytes.length - 1;
		for ( int offset = 0; offset < 2; ++offset ) {
			final int cut = Math.max( -1, b - 8 );
			long u = 0;
			for ( int p = 0; b > cut; --b, p += 8 ) {
				u |= ( bytes[ b ] & 0xffL ) << p;
			}
			dataAccess.setValue( k + offset, u );
		}
	}
```
Could you try whether that works?

awalter17:
@tpietzsch Thanks for the suggestion, your shortened method works perfectly! 
I also added in a regression test that confirms this, and a regression test that confirms the ClassCastException issue is fixed.

tpietzsch:
@awalter17 thanks a lot!

PR #98, opened 2015-08-07
IterableIntervalProjector2D: Target should use X,Y not dimX, dimY

dietzc:
DimX and dimY are the dimensions representing the plane of the
RandomAccessible to be rendered. Any of these can be > 1. Target
is always two dimensional and therefore should use X,Y to access
the plane dimensions.

Sorry for the typo in the branch-name.


dietzc:
@axtimwalde @StephanPreibisch any concerns? I'm pretty confident that this makes sense.

axtimwalde:
@dietzc Your code, you probably know what you're doing.  I would, though, remove the constants X and Y and write 0 and 1 directly (we enumerate coordinates by numbers throughout ImgLib2) or at least make them static.

dietzc:
makes sense. I will fix this and merge

PR #106, opened 2015-11-11
PlanarSubsetCursor: Bug in calculation of size of plane

dietzc:
see http://forum.imagej.net/t/cursor-getting-lost-inside-an-interval-with-only-one-voxel/312/4


tpietzsch:
cool, thanks @dietzc !

Issue #105, opened 2015-10-22
Add get/setBigInteger to IntegerType?


awalter17:
Hello! 

Currently the [IntegerType](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/type/numeric/IntegerType.java) interface has getters and setters only for `long` and `int` values. However, this interface is implemented by [Unsigned128BitType](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/type/numeric/integer/Unsigned128BitType.java) which is backed by a `BigInteger`. This causes an issue if someone ever wants to get the value of an IntegerType. 

```
public <T extends IntegerType<T>> void printValue(final T input) {
    System.out.println(input.getIntegerLong());
}
```

For example, in the above method if `input` turned out to be an Unsigned128BitType with a value greater than 2<sup>63</sup> - 1, I could never get the actual value of that Unsigned128BitType without knowing it is an Unsigned128BitType. 

As such, I think it would be beneficial to add the methods `getBigInteger()` and `setBigInteger()`, to the IntegerType interface. 

Please let me know your thoughts on this, and thank you in advance for your input.

tpietzsch:
@awalter17 I agree. It would be good to add those

ctrueden:
Addressed by #109.

Issue #102, opened 2015-10-08
How to fix UnsignedLongType?


awalter17:
Hello all!

I noticed several issues with [UnsignedLongType](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/type/numeric/integer/UnsignedLongType.java). 
1. getMaxValue( ) is 0xffffffffL and it should be 0xffffffffffffffffL (this value cannot be represented in java as an unsigned value)
   - Maybe we could do something similar to what is done in [Unsigned128BitType](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/type/numeric/integer/Unsigned128BitType.java)?
2. compare( ) doesn't work for negative values
   - Can't compare unsigned `long`s larger than 2<sup>63</sup> - 1
   - Since there is no larger primitive than long, it uses long
     - this is also inconsistent with how the other unsigned types implement compare
3. get( ) returns a long
   - Calling get( ) on an UnsignedLongType can then return a negative value, if it was constructed with a negative 
   - Maybe it should return a `BigInteger` as to return unsigned `long`s larger than 2<sup>63</sup> - 1
   - Inconsistent with other unsigned types which return the next larger primitive (which can't be done here)

I think one possible way to fix some of these issues would be to utilize `BigInteger` similar as to what is done in [Unsigned128BitType](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/type/numeric/integer/Unsigned128BitType.java). However, I am not certain if this is the best way to solve these issues. 

Please let me know your thoughts on this, or if anything I said was unclear.

tpietzsch:
good catch @awalter17!

Regarding `compare()`, something like this should work:
```
	static int compare( final long a, final long b )
	{
		return a < 0 ?
				( b < 0 ? ( int ) ( a - b ) : 1 ) :
				( b < 0 ? -1 : ( int ) ( a - b ) );
	}
```
The logic is sound, I tried it with bytes:
```
	static int compare( final byte a, final byte b )
	{
		return a < 0 ?
				( b < 0 ? a - b : 1 ) :
				( b < 0 ? -1 : a - b );
	}

	public static void main( final String[] args )
	{
		for ( int a = 0; a <= 255; ++a )
		{
			for ( int b = 0; b <= 255; ++b )
			{
				final int i = compare( ( byte ) a, ( byte ) b );
				if ( i == 0 )
				{
					if ( a != b )
[...]

PR #109, opened 2015-11-12
Fix unsigned long type

awalter17:
Hello!

These are the fixes for UnsignedLongType, as mentioned in #102 . That being said, upon further inspection I noticed that the compareTo method was fine. So I left the compareTo method alone, but I did keep the tests I made for it.

Additionally on this branch I added the methods getBigInteger, and setBigInteger to the IntegerType interface. This was done to allow for n-bit integer arithmetic, as well as the issue mentioned in #105 . The only problem I had with implementing these, was that [UnboundedIntegerType](https://github.com/imagej/imagej-common/blob/master/src/main/java/net/imagej/types/UnboundedIntegerType.java) in ImageJ-Common implements IntegerType.

Lastly, I also fixed a masking issue with 64 bits in UnsignedVariableBitLengthType.

Please let me know your thoughts on these changes or if you have any questions. Thank you in advance for your feedback!


ctrueden:
@tpietzsch OK to merge this?

tpietzsch:
@awalter @ctrueden Yes, looks great. Thanks!

ctrueden:
:+1:

PR #111, opened 2016-01-13
Fix equals for integer bit types

ctrueden:
This improves behavior of equals for types extending `AbstractIntegerBitType`:
`Unsigned2BitType`, `Unsigned4BitType` and `Unsigned12BitType`.

The reason it is necessary is because AbstractIntegerBitType extends
AbstractBitType rather than AbstractRealType, and hence does not inherit the
latter's equals, hashCode or compareTo method implementations. (The equals
implementation here is actual equal to the one in AbstractRealType.)

@tpietzsch Does this look OK?


tpietzsch:
looks great, thanks @ctrueden

PR #114, opened 2016-01-17
Adds getters to several classes

dietzc:
allows decomposition (and subsequent recomposition) of the wrapped objects. Useful for serialization. Potentially we want `WrappedSource`, `WrappedInterpolationFactory,... interfaces later.


axtimwalde:
I am not completely sure why I called the source target.  Now that it gets a getter method that seems like the right time to avoid writing this in stone and make it a thing if it was just stupid.  Comments?

PR #116, opened 2016-01-19
Mark lightweight data objects with a common iface

ctrueden:
The concept of a "view" is in several places across ImgLib2; it is convenient to have them all share a common interface.

Firstly, this allows for runtime introspection, so that programs can decide whether it is warranted to make a copy of the data (i.e., to "burn it in"), to avoid:

1. Costly recomputation when the lightweight data is accessed; and
2. Corrupting underlying data when the lightweight data mutates.

Secondly, this interface provides a common point for exploring the ImgLib2 API, so that programmers can easily browse which lightweight data structures exist, and how they are implemented.

### DIFF
--- a/src/main/java/net/imglib2/View.java
+++ b/src/main/java/net/imglib2/View.java
@@ -32,61 +32,28 @@
  * #L%
  */
 
-package net.imglib2.interpolation;
-
-import net.imglib2.EuclideanSpace;
-import net.imglib2.RealInterval;
-import net.imglib2.RealRandomAccess;
-import net.imglib2.RealRandomAccessible;
-import net.imglib2.View;
+package net.imglib2;
 
 /**
- * A {@link RealRandomAccessible} that is generated through interpolation.
+ * An interface which marks an object that is a lightweight wrapper or "view"
+ * around actual data.
+ * <p>
+ * This is useful to introspect at runtime whether a data object might benefit
+ * from being copied to "burn in" the information, for reasons such as:
+ * </p>
+ * <ul>
+ * <li>To avoid costly recomputation, at the expense of more memory use.</li>
+ * <li>To avoid corrupting the wrapped data structure when the object is
+ * mutated.</li>
+ * </ul>
  * 
- * @author ImgLib2 developers
- * @author Stephan Saalfeld <saalfeld@mpi-cbg.de>
+ * @author Curtis Rueden
  * @author Tobias Pietzsch
+ * @author Christian Dietz (University of Konstanz)
+ * @see net.imglib2.view.Views
+ * @see net.imglib2.converter.Converters
  */
-final public class Interpolant< T, F extends EuclideanSpace > implements RealRandomAccessible< T >, View
+public interface View
 {
-	final protected F source;
-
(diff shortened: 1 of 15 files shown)


ctrueden:
@tpietzsch Here is my first cut. I am not sure about certain types:

* Should `ConvertedRandomAccess` and `ConvertedCursor` be tagged? As the supporting objects, I am not sure it is needed—we never pass around `RandomAccess` or `Cursor` directly in Ops.
* I did not tag `AbstractWrappedInterval`, since I guess some wrapped intervals should not be considered this sort of "lightweight" data object...
* But I _did_ tag `IterableRandomAccessibleInterval`, which I am not sure is correct.
* I mostly only tagged `net.imglib2.converter` and `net.imglib2.view`. Also `net.imglib2.view.IntervalView`. Are there other packages that should be included?

ctrueden:
@axtimwalde Once this is finalized and released, I will file another PR for `imglib2-realtransform` for the objects which `RealViews` generates.

tpietzsch:
@ctrueden
> Should ConvertedRandomAccess and ConvertedCursor be tagged? As the supporting objects, I am not sure it is needed—we never pass around RandomAccess or Cursor directly in Ops.

I think `ConvertedRandomAccess` and `ConvertedCursor` should not be tagged, for now.
I would use `View` for the Accessibles, not the Accessors. If needed we can add something later.

> I did not tag AbstractWrappedInterval, since I guess some wrapped intervals should not be considered this sort of "lightweight" data object...

Yes.

> But I did tag IterableRandomAccessibleInterval, which I am not sure is correct.

I think that is correct. @axtimwalde do you agree?

axtimwalde:
I wouldn't tag the accesses either, but my feelings about this interface are not very sharp.  There is a point in tagging the accesses for another purpose that is not covered by this tag: Do modifications write through, and where to?  This may be interesting to decide whether e.g. painting into a View would result in modified pixels.  We currently do not have such a tag and I believe that it is not possible to get this information by tagging.  It is even worse, e.g. Interpolant can both write through or not write through depending on which Interpolator is used.  Here, even tagging the Access wouldn't help because it transparently uses an Interpolator.  Sorry for hijacking this discussion, I just noticed that the information that can be extracted from the View tag is rather limited, may be I am too negative here.

tpietzsch:
@axtimwalde Yes, I agree. The only use-case at the moment is the one that @ctrueden indicated: Act as a hint that you (that is, imagej-ops) might want to make a copy of the Accessible because pixel access may be costly (and you expect to access pixels repeatedly). It doesn't say anything about how costly accesses are etc, so it carries little information. I think it doesn't hurt to have it and if it doesn't turn out to be useful, I think we can remove it again without causing much trouble.

We can think about providing much more detailed information, like what you outlined regarding write-through. That seems like a rather big project however.

ctrueden:
Thanks for the discussion! Sounds like this is good to merge, then?

tpietzsch:
From my point of view, yes, it can be merged. I'm not sure whether there are any other classes that should be tagged `View`, but we can add more later.

ctrueden:
OK, great!

PR #117, opened 2016-01-20
Accessor-method ('getTarget') in RandomAccessibleOnRealRandomAccessible.

hornm:
Another accessor method ...


axtimwalde:
@tpietzsch That should be "source" too---right?

tpietzsch:
@axtimwalde yes, right. I'll change that

PR #115, opened 2016-01-18
Fix  PlanarSubsetLocalzingCursor

dietzc:
position calculation was mixed between subset interval and global container.

PR #119, opened 2016-01-29
Debug ConnectedComponents

jaywarrick:
Seems to work now but certainly should be reviewed by someone who knows
what they are doing in this class. Line 155 and 302 are edits from
@dietzc while line 297 is from me.


dietzc:
index container has to be calculated differently. see https://github.com/imglib/imglib2/pull/115.
your implementation works for your certain use-case but will fail in the general one, as far as I can see.

thanks! this helped debugging

Issue #110, opened 2016-01-02
LongType.setReal(longNumber) sets the variable to Integer.MAX_VALUE


LeonYang5114:
When setting the real part of a LongType to a large number using setReal(), the value of that variable will be set to Integer.MAX_VALUE instead.
Code to generate the situation:
```
LongType num = new LongType();
num.setReal(9876543210L);    // number larger than 2147483647
System.out.println(num.get());    // outputs "2.147483647E9"
```
The issue is rooted in the fact that `setReal()` accepts either `float` or `double` as input, but when using a `long`, Java automatically selects the [`float` version](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/type/numeric/integer/LongType.java#L177). After the `long` is rounded as a `float`, it is then [casted to an `int`](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/util/Util.java#L487), which causes the problem.
We could just cast the `long` to a `double` manually to avoid this problem, but I think that's quite tricky. Removing the `setReal(float)`method could be a solution.

tpietzsch:
@LeonYang5114 No, we cannot remove the `setReal(float)` method. It is defined further up in the hierarchy and for example for `FloatType` it makes a lot of sense. Any reason why you wouldn't use `setInteger()` instead?

ctrueden:
> Any reason why you wouldn't use `setInteger()` instead?
If you have a `T extends RealType<T>` then `setInteger` isn't available. I think it is OK to write `num.setReal((double) myLong)`. Alternately, and perhaps better, would be to add a `setReal(long)` to the API contract of `RealType` so that specific type implementations can do something less lossy when it comes to `long`s.

tpietzsch:
@ctrueden Yep, that is a good idea

ctrueden:
Cool, @LeonYang5114 would you care to make a PR for that?

PR #120, opened 2016-01-30
Create API typed on long for ComplexType

LeonYang5114:
getRealLong, getImginaryLong, setReal(long), setImginary(long) are created in the ComplexType interface, and its implementations are modified accordingly.

This branch addresses issue #110 . To be more clear, that issue points out the unexpected behavior:
```
public <T extends RealType<T>> void setLargeLongNum(T final num) {
    num.setReal(9876543210L);    // number greater than Integer.MAX_VALUE
}
...
LongType num = new LongType();
setLargeLongNum(num);
System.out.println(num.get());    // outputs "2.147483647E9"
...
```

After looking into the source code, it seems to me that it makes sense to call `num.setReal((double) 9876543210L);` instead of the call above, since `double` and `long` both have same length of storage. But users might intuitively think that, although converting `long` to `float` is quite lossy, we should at least get something close to the original number (9876543210L) instead of `Integer.MAX_VALUE`. So should this be considered a bug? or just that it is inherently tricky? Probably this branch is not that necessary.

In addition, I did not create `ComplexLongType`, since I am not sure if we need that, but it would be pretty easy to create. I also did not create tests for this feature, since I think the their `float` and `double` counterparts do not have tests either.

### DIFF
--- a/src/main/java/net/imglib2/type/numeric/ComplexType.java
+++ b/src/main/java/net/imglib2/type/numeric/ComplexType.java
@@ -52,13 +52,25 @@ public interface ComplexType< T extends ComplexType< T >> extends NumericType< T
 
 	public void setReal( double f );
 
+	public void setReal( int f );
+
+	public void setReal( long f );
+
 	public void setImaginary( float f );
 
 	public void setImaginary( double f );
 
+	public void setImaginary( int f );
+
+	public void setImaginary( long f );
+
 	public void setComplexNumber( float r, float i );
 
 	public void setComplexNumber( double r, double i );
+
+	public void setComplexNumber( int r, int i );
+
+	public void setComplexNumber( long r, long i );
 
 	public float getPowerFloat();
 
--- a/src/main/java/net/imglib2/type/numeric/complex/AbstractComplexType.java
+++ b/src/main/java/net/imglib2/type/numeric/complex/AbstractComplexType.java
@@ -169,6 +169,20 @@ public abstract class AbstractComplexType< T extends AbstractComplexType< T >> i
 	}
 
 	@Override
+	public void setComplexNumber( final int r, final int i)
+	{
+		setReal( r );
+		setImaginary( i );
+	}
+
+	@Override
+	public void setComplexNumber( final long r, final long i)
+	{
+		setReal( r );
+		setImaginary( i );
+	}
+
+	@Override
 	public boolean equals( final Object o )
 	{
 		if ( !getClass().isInstance(o) )
--- a/src/main/java/net/imglib2/type/numeric/complex/ComplexDoubleType.java
+++ b/src/main/java/net/imglib2/type/numeric/complex/ComplexDoubleType.java
@@ -150,6 +150,18 @@ public class ComplexDoubleType extends AbstractComplexType< ComplexDoubleType > 
 	}
 
 	@Override
+	public void setReal( final int r )
+	{
+		dataAccess.setValue( realI, r );
+	}
+
+	@Override
+	public void setReal( final long r )
+	{
+		dataAccess.setValue( realI, r );
+	}
+
+	@Override
 	public void setImaginary( final float i )
 	{
 		dataAccess.setValue( imaginaryI, i );
@@ -157,6 +169,18 @@ public class ComplexDoubleType extends AbstractComplexType< ComplexDoubleType > 
 
 	@Override
 	public void setImaginary( final double i )
+	{
+		dataAccess.setValue( imaginaryI, i );
+	}
+
+	@Override
+	public void setImaginary( final int i )
+	{
+		dataAccess.setValue( imaginaryI, i );
+	}
+
+	@Override
+	public void setImaginary( final long i )
 	{
 		dataAccess.setValue( imaginaryI, i );
 	}
--- a/src/main/java/net/imglib2/type/numeric/complex/ComplexFloatType.java
+++ b/src/main/java/net/imglib2/type/numeric/complex/ComplexFloatType.java
@@ -150,6 +150,18 @@ public class ComplexFloatType extends AbstractComplexType< ComplexFloatType > im
 	}
 
 	@Override
+	public void setReal( final int r )
+	{
+		dataAccess.setValue( realI, r );
+	}
+
+	@Override
+	public void setReal( final long r )
+	{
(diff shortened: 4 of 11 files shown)


tpietzsch:
@LeonYang5114 @ctrueden I don't like the addition of new `getRealLong, getImginaryLong` methods. I think this is dangerous. If you want to read as integers you should have to cast explicitly. Adding overloaded `setReal, setImaginary` is fine, but read methods are a different topic. Also: did you check what happens if you now call `setReal()` with an `int`? I think previously it would call the `setReal(float)` variant, now possibly the `long` variant. Could this lead to changes in behaviour / runtime? Maybe it would be better to add explicit `setReal(int)` as well?

ctrueden:
@tpietzsch I do not understand why `getRealLong` / `getImaginaryLong` is dangerous. In fact, it is necessary to avoid data loss in some cases. I see `double` and `long` as Java's two foremost "richest" primitive numerical types, each with its pros and cons. So why not give them equal standing in the API? If the only type-agnostic way to extract a `long` is to write `(long) getRealDouble()` then for types such as `LongType` and `Unsigned128BitType`, the value you obtain will be "lossier" than if we have a dedicated `getRealLong()` method which each type implements to do the best thing possible for itself.

ctrueden:
@LeonYang5114 To play devil's advocate: the `IntegerType` interface already has `getIntegerLong()` and `getBigInteger()`. Maybe this is good enough?

LeonYang5114:
I just read about [this](http://stackoverflow.com/questions/1293819/why-does-java-implicitly-without-cast-convert-a-long-to-a-float). I think Java as a language is designed to compromise somehow, so it has some lossiness (e.g. convert `long` to `float` instead of `double` by default). It seems to be our responsibility to do that explicit casting from `long` to `double` (to reduce lossiness) if we want.

For the other part, I think that if we want absolutely no lossiness when passing `long` variable, we should use `LongType` explicitly. If we use `RealType`, we have to be lossy. It seems fair game to me. If we only have `setLongXXX` and without `getLongXXX`, the overall effect would still be lossy, right?

At this point of time, I suggest that we make no changes, so we should cast explicitly a `long` to a `double` in the above example, or make the [utility method](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/util/Util.java#L485-L488) that rounds a `float` to return a `long`, instead of returning an `int`.

I will make the changes as indicated by @tpietzsch if you guys like that `long/int` API. I am not sure if we need that now.

tpietzsch:
@ctrueden You have a point with the potential lossiness of not having `getRealLong()`. My concern is about lossiness if you have an actual `DoubleType` that might carry values between 0 and 1. Then the presence of `getRealLong()` might lure users into thinking that it will always return sensible values. I think there is no good general answer. If you want to do arithmetic in `long` or `float` or any other type, the only way in which you can be sure that it "works" is if you know more about the actual type. Otherwise, the safer way to do it is using the arithmetic operations of the `Type`.

@ctrueden, @LeonYang5114 Can you give more details on the where you want to use this and why it is not an option to stay within the type algebra?

@axtimwalde @StephanPreibisch What is your opinion on this?

axtimwalde:
I strongly vote against this pull request.  Please do not add `getRealLong()`, `getRealInteger()`, `getImaginaryLong()` and `getImageinaryInteger()` to `ComplexType`, nor to `RealType`.  The current type hierarchy makes sense in the reading direction in that it expands based on increasing specialization of the type from an algebraic viewpoint, precision is not considered.  The current type hierarchy makes no sense in the writing direction where all that is pure nonsense.  We have discussed separating the reading and writing hierarchies into independent interfaces several times but since this came after premature release of ImgLib2, we have been hesitant to actually do it.  Keep in mind that ImgLib2 types at that interface level make no guarantees about the precision that is preserved.   Doing calculations on RealType or ComplexType typically makes no sense because of this limitation, but you can typically assume to be able to read their data into double precision which is implicitly considered optimal precision.  Introducing integer getters in real types makes this worse, not better because they allow no such assumption, i.e. they are not only useless but add violate the sole reliable principle in this framework.  Adding the setters does no harm other than potential unintended change in behavior as @tpietzsch already pointed out.  That, however, means that you would have to carefully test all existing code depending on ImgLib2 that you know of to not disrupt those people's valuable developer time, announce the changes and potential harm with emphasis, ready to help remedy unintended consequences, and increase the major version number in the release.  I have done this recently with the mpicbg library for a very very good reason and it was a major pain in the neck.  I do not know if that's worth it. Honestly, I think if we wish to undergo this pain, we should overhaul the type hierarchy based on what we've learned about separating the hierarchies of algebraic reading, algebraic writing, reading precision, and writing precision.

StephanPreibisch:
I agree with @axtimwalde. If it is of use, we could add a a ComplexIntegerType to allow integer computation with complex numbers staying consistent with our current read based hierarchy.

ctrueden:
OK, I closed it.

> double precision which is implicitly considered optimal precision

This makes no sense to me. Double is no more "optimal" than long. Both can be lossy given the generality of these types.

> Doing calculations on RealType or ComplexType typically makes no sense because of this limitation

I agree. Unfortunately, nearly every algorithm I have seen coded in ImgLib2 uses `getRealDouble` and `setReal`, despite it being problematic for these reasons. Fixing this will take time as we introduce the many necessary math ops built on add/sub/mull/div rather than Java's math operators.

axtimwalde:
>         double precision which is implicitly considered optimal
>         precision
>         
> 
> This makes no sense to me. Double is no more "optimal" than long. Both
> can be lossy given the generality of these types.
> 
Fully agree, but the only implicit assumption in the existing hierarchy
is that getReal() is the best thing you can do.  This is correct from
the algebraic position that does not consider precision (because an
integer is a real value).  The current hierarchy ignores precision at
all levels.  There is no coherent precision based hierarchy, neither in
reading nor writing direction.
>

ctrueden:
> the only implicit assumption in the existing hierarchy is that getReal() is the best thing you can do.  This is correct from the algebraic position that does not consider precision (because an integer is a real value).  The current hierarchy ignores precision at all levels.

Then the method should have been called `getReal()`, not `getRealDouble()`. And `getRealFloat()` should not exist.

tpietzsch:
That is probably correct. I would be ok with deprecating `getRealFloat()` and removing it with the next major version.

PR #123, opened 2016-03-06
Add convenience method to extend a RAI and define an Interval on it again

dietzc:
see commit message.

### DIFF
--- a/src/main/java/net/imglib2/view/ExtendedRandomAccessibleInterval.java
+++ b/src/main/java/net/imglib2/view/ExtendedRandomAccessibleInterval.java
@@ -35,10 +35,12 @@
 package net.imglib2.view;
 
 import net.imglib2.Interval;
+import net.imglib2.Positionable;
 import net.imglib2.RandomAccess;
 import net.imglib2.RandomAccessible;
 import net.imglib2.RandomAccessibleInterval;
 import net.imglib2.View;
+import net.imglib2.RealPositionable;
 import net.imglib2.outofbounds.OutOfBounds;
 import net.imglib2.outofbounds.OutOfBoundsFactory;
 import net.imglib2.util.Intervals;
@@ -52,7 +54,7 @@ import net.imglib2.util.Intervals;
  * @author Stephan Saalfeld <saalfeld@mpi-cbg.de>
  * @author Tobias Pietzsch
  */
-public final class ExtendedRandomAccessibleInterval< T, F extends RandomAccessibleInterval< T > > implements RandomAccessible< T >, View
+public final class ExtendedRandomAccessibleInterval< T, F extends RandomAccessibleInterval< T > > implements RandomAccessibleInterval< T >, View
 {
 	final protected F source;
 
@@ -93,5 +95,89 @@ public final class ExtendedRandomAccessibleInterval< T, F extends RandomAccessib
 
 		if ( Intervals.contains( source, interval ) ) { return source.randomAccess( interval ); }
 		return randomAccess();
+	}
+
+	@Override
+	public long min( int d )
+	{
+		return source.min( d );
+	}
+
+	@Override
+	public void min( long[] min )
+	{
+		source.min( min );
+	}
+
+	@Override
+	public void min( Positionable min )
+	{
+		source.min( min );
+	}
+
+	@Override
+	public long max( int d )
+	{
+		return source.max( d );
+	}
+
+	@Override
+	public void max( long[] max )
+	{
+		source.max( max );
+	}
+
+	@Override
+	public void max( Positionable max )
+	{
+		source.max( max );
+	}
+
+	@Override
+	public double realMin( int d )
+	{
+		return source.realMin( d );
+	}
+
+	@Override
+	public void realMin( double[] min )
+	{
+		source.realMin( min );
+	}
+
+	@Override
+	public void realMin( RealPositionable min )
+	{
+		source.realMin( min );
+	}
+
+	@Override
+	public double realMax( int d )
+	{
+		return source.realMax( d );
+	}
+
+	@Override
+	public void realMax( double[] max )
+	{
+		source.realMax( max );
+	}
+
+	@Override
+	public void realMax( RealPositionable max )
+	{
+		source.realMax( max );
+	}
+
+	@Override
+	public void dimensions( long[] dimensions )
+	{
+		source.dimensions( dimensions );
+	}
+
+	@Override
+	public long dimension( int d )
+	{
+		return source.dimension( d );
 	}
 }
--- a/src/main/java/net/imglib2/view/ExtendedRandomAccessibleInterval.java
+++ b/src/main/java/net/imglib2/view/ExtendedRandomAccessibleInterval.java
@@ -52,7 +52,7 @@ import net.imglib2.util.Intervals;
  * @author Stephan Saalfeld <saalfeld@mpi-cbg.de>
  * @author Tobias Pietzsch
  */
-final public class ExtendedRandomAccessibleInterval< T, F extends RandomAccessibleInterval< T > > implements RandomAccessible< T >, View
+public final class ExtendedRandomAccessibleInterval< T, F extends RandomAccessibleInterval< T > > implements RandomAccessible< T >, View
 {
 	final protected F source;
 
--- a/src/main/java/net/imglib2/view/ExtendedRandomAccessibleInterval.java
+++ b/src/main/java/net/imglib2/view/ExtendedRandomAccessibleInterval.java
@@ -64,6 +64,16 @@ final public class ExtendedRandomAccessibleInterval< T, F extends RandomAccessib
 		this.factory = factory;
 	}
 
+	public F getSource()
+	{
+		return source;
+	}
+
+	public OutOfBoundsFactory< T, ? super F > getOutOfBoundsFactory()
+	{
+		return factory;
+	}
+
 	@Override
 	final public int numDimensions()
 	{
@@ -83,15 +93,5 @@ final public class ExtendedRandomAccessibleInterval< T, F extends RandomAccessib
 
 		if ( Intervals.contains( source, interval ) ) { return source.randomAccess( interval ); }
 		return randomAccess();
-	}
-
-	public F getSource()
-	{
-		return source;
-	}
-	
-	public OutOfBoundsFactory< T, ? super F > getOutOfBoundsFactory()
-	{
-		return factory;
 	}
 }


Squareys:
@tpietzsch :+1: this would be very useful, indeed.

@dietzc regarding naming I would suggest something like `extendKeepInterval`, or `extendOutside[s]`? Just an idea, `extendAndWrap` is fine, too. `:)`

stelfrich:
We had introduced that functionality in [`RAIs.extend()`](https://github.com/imagej/imagej-ops/blob/353ab75f54f1edb439ec51c1c27d21d232000cc9/src/main/java/net/imagej/ops/special/chain/RAIs.java#L92) in Ops. I agree, however, that we should move it here :+1:

tpietzsch:
I agree that this is useful functionality, but is it really necessary to add a method for this? I mean it's just a few characters more to type. And I'm afraid that next you'll want to add `extendMirrorSingleAndWrap`, `extendBorderAndWrap` etc... I think it's more confusing than helpful to add these.

However, I'm fine to add those if people prefer it. Any opinions on this, @ctrueden @axtimwalde @StephanPreibisch ?
We need to find a good name. I don't like `extendAndWrap`. So far I like `extendOutside` the best. Any other suggestions?

dietzc:
@tpietzsch the method doesn't need to live in `Views`. I had the same concerns wuith `Views.extendBorderAndWrap` etc. Any other suggestions? I want to avoid that we introduce such helpers in `imagej-ops`, because this is really `imglib2` functionality (see https://github.com/imagej/imagej-ops/pull/364). About naming: `extendOutside` sounds good.

ctrueden:
Instead of adding a new method: why can't `ExtendedRandomAccessibleInterval` simply implement `Interval`, using its source?

axtimwalde:
@ctrueden I think that is a good idea and I will support a pull request :).  @dietzc @tpietzsch @stelfrich @Squareys I do not think that adding the proposed method makes sense as it saves a low count of characters by cluttering the API.  Regarding imglib2 related helper methods in imagej-ops: I believe that there will be more to come and so discussing a good place in imagej-ops is probably a good idea.

ctrueden:
I took the liberty of force updating this branch. @dietzc I hope you do not mind.

tpietzsch:
@ctruede @axtimwalde. No, that is a bad idea!

ctrueden:
@tpietzsch Why?

tpietzsch:
@ctrueden Sorry, I meant to explain, just wanted to quickly intervene, before someone merges it to master...

ctrueden:
> just wanted to quickly intervene, before someone merges it to master...

Understood.

> I meant to explain

OK, great. I feel pretty confident that this is a good change, so we can have a robust discussion about it here. :grin:

tpietzsch:
The potential reason is this:
https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/RandomAccessible.java#L60-L104

That was a decision we made when deciding how the `Views` simplification should work (it might not have been the best decision). Now the question is how the `randomAccess()` of the modified `ExtendedRandomAccessibleInterval` should work. If it follows the rule set up in the above javadoc comment, it will break existing code, which relies on the fact that the `randomAccess()` of the extended interval is defined everywhere. If it breaks the rule set up in the javadoc it leads to inconsistent behaviour across different `RandomAccessible`s. For example if you have a `ExtendedRandomAccessibleInterval i` of the modified kind you propose, then `i.randomAccess()` will be defined out of bounds while `Views.interval(i, i).randomAccess()` will not be defined (although they both represent exactly the same region of the image.

It may be time to rethink those rules, but let's not rush into it. A lot of code would need to be reviewed for potential damage...

tpietzsch:
Coming back to the original PR, I changed my mind a bit. I think it is not so terrible to add the new methods, although they are so short. It would allow easier construction of view chains without intermediate variables.

I.e., to do
```
Views.extendBorderAndWrap( Views.someviewchain(...) )
```
now you need to have a local variable
```
a = Views.someviewchain(...);
Views.interval( Views.extendBorder( a ), a )
```
because you need the `a` twice.

tpietzsch:
I'm torn. Maybe the inconsistent behaviour is ok. Difficult.

axtimwalde:
@tpietzsch no worry, I was not about to merge this in a rush ;), but I can feel the fear with you.  Here is another point: If it is not about breaking the contract in
https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/RandomAccessible.java#L60-L104
what then is the purpose of this method anyways?  I can see how it is useful to be able to get a `RandomAccess` from a `RandomAccessibleInterval` and then assume that it is defined outside the boundaries but that idea breaks the above concept.  The 'consuming' code in such a situation should deal with a `RandomAccessible` and a separate `Interval`, so I think that we either (a) break the contract completely (by making `ExtendedRandomAccessibleInterval extend Interval` or (b) not facilitate this concept at all, keeping everything as is (as it is easy enough to still do what one wants, even if misguided).  I do not like this new method for the above reasons.  Good counterarguments always welcome :).

ctrueden:
Thanks for the explanation, @tpietzsch. I didn't realize that by adding the interval back via `Views.interval`, the `RandomAccess` would again be limited only to the `Interval`. In that case, I'm actually curious why any ops need to use this pattern—i.e., what is even the point of rebounding the interval if the random access to the outside is "lost" again.

> It may be time to rethink those rules, but let's not rush into it. A lot of code would need to be reviewed for potential damage...

I would certainly say the rules should be reconsidered. There is definitely a distinction between "a bounded interval that cannot do RA outside its bounds" and "a bounded interval which nonetheless can do RA outside its bounds somehow." From my perspective this distinction should somehow be modeled using distinct types. Then Views can "do the right thing" in both cases, maybe?

tpietzsch:
@axtimwalde Basically, the construct makes it clear that the original RAI is extended if you explicitly request it. In the code that you would pass the extended and cropped RAI (let's call it `ecrai`) into:

If you call `ecrai.randomAccess()` then the whole thing is indeed pointless. You will get an access on the original RAI that fails if you access out of bounds.

If you call `ecrai.randomAccess(interval)`, specifying the `interval` in which you intend to use it, then ECRAI will do it's best to give you an access that is valid in the specified interval. The out-of-bounds extension will kick in if the interval goes outside the original RAI.

This is useful for example in the following case: You have a method that runs convolution filter on an image producing an image of the same size. Now, if you inside the method cut the image into blocks to parallelize the computation, you will get fast non-out-of-bounds accesses if the block + boundary is inside the original RAI. For the "border' blocks you will get out-of-bounds accesses.

An alternative consistent policy would be:
`randomAccess()` always gives you an access that is defined as far as possible.
Only if you call `randomAccess(interval)` you may get an access that is invalid outside the defined `interval`.
I cannot fully grasp what this would mean for Views, i.e., if there is anywhere a problem where something just cannot work anymore.
For sure, some existing code will slow down because it relies on Views stripping away the out-of-bounds if `Views.interval()` allows for it. Basically we would need go through all algorithms and switch from calling `randomAccess()` by default to `randomAccess(interval)` by default.

Finally, another policy would be:
If you call `randomAccess()`, all bets are of.
If you want to be sure, use `randomAccess(interval)`.
With this, we can make the proposed ExtendedRandomAccessibleInterval change. I don't like it very much because it is not as definite as the alternatives. But supposedly very little existing code would break (although I'm not sure of that).

tpietzsch:
> An alternative consistent policy would be:
> randomAccess() always gives you an access that is defined as far as possible.
> Only if you call randomAccess(interval) you may get an access that is invalid outside the defined interval.
> I cannot fully grasp what this would mean for Views, i.e., if there is anywhere a problem where something just cannot work anymore.
> For sure, some existing code will slow down because it relies on Views stripping away the out-of-bounds if Views.interval() allows for it. Basically we would need go through all algorithms and switch from calling randomAccess() by default to randomAccess(interval) by default.

Basically it boils down to:
Now if you want to be sure to get the fastest access for a given interval you can do
```
Views.interval( accessible, interval ).randomAccess()
```
and there is code that relies on it.
With the changed policy you would need to
```
accessible.randomAccess(interval)
```

Perhaps more importantly, if you want to iterate a sub-interval, now you do
```
Views.interval( accessible, interval ).cursor()
```
and if the cursor is based on wrapping a `RandomAccess`, that `RandomAccess` will be the fastest possible.
With the changed policy: out of the box you would get a cursor backed by a slow `RandomAccess`.
We would need to change for example `RandomAccessibleIntervalCursor` to use `randomAccess(interval)` to get the speed back. And I suspect there are several such non-obvious places that would need to be revised.

tpietzsch:
> If you call ecrai.randomAccess(interval), specifying the interval in which you intend to use it, then ECRAI will do it's best to give you an access that is valid in the specified interval. The out-of-bounds extension will kick in if the interval goes outside the original RAI.

Or, of course, equivalently you can do `Views.interval(ecrai, interval).randomAccess()` with a larger interval (e.g. `interval=Intervals.expand(ecrai,border)`) to get out-of-bounds if it is available. I believe that is the way it is actually used in some algorithms.

tpietzsch:
> With the changed policy: out of the box you would get a cursor backed by a slow RandomAccess.
> We would need to change for example RandomAccessibleIntervalCursor to use randomAccess(interval) to get the speed back. And I suspect there are several such non-obvious places that would need to be revised.

And this would even be the case with just leaving everything as is, just making the proposed change to `ExtendedRandomAccessibleInterval`. The cursor on any interval of that thing might be slow, while with the current design, `Views` can look through the `ExtendedRandomAccessibleInterval` and for example give you directly a cursor (*not* wrapping a `RandomAccess` *at all*) of the underlying `ArrayImg` if the interval happens to be the right one.

So, you see: The implications of this change are potentially far-reaching. If we want to do this, I would definitely post-pone to a hackathon where several people work on it dedicatedly.

LeonYang5114:
Is it possible to just rename `ExtendedRandomAccessibleInterval` to `ExtendedRandomAccess` and make the `ExtendedRandomAccessibleInterval` as proposed, i.e. a `RAI` that has out-of-bound RA ability. Then no code will break, and the proposed class will be a handy wrapper for an `ExtendedRandomAccessible` and an `Interval`.

The main reason I want this happen is that I hope we rely on the compiler to check if out-of-bound RA is allowed for a specific `RAI`, and fail early if not. If we type on the original `ExtendedRandomAccessibleInterval`, then the `Interval` information is missing. A wrapper would be very convenient.

For the behavior of the modified `ExtendedRandomAccessibleInterval`, let's call it `erai`. I think we could assume that `erai.randomAccess()` has the full out-of-bound ability. Then we are concerned about whether the following should be out-of-bound RA, and how should they be optimized:

1. `Views.interval(erai, fullyContainedInterval).randomAccess()`
2. `Views.interval(erai, notFullyContainedInterval).randomAccess()`
3. `erai.randomAccess(fullyContainedInterval)`
4. `erai.randomAccess(notFullyContainedInterval)`

Intuitively, I think 1 and 2 above should not be out-of-bound RA. I am not sure for how 3 and 4 should work. Do you think this idea has any potential for further discussion?

tpietzsch:
@LeonYang5114 No, I don't think that has much potential.

> The main reason I want this happen is that I hope we rely on the compiler to check if out-of-bound RA is allowed for a specific RAI, and fail early if not.

I think it is not useful to let the compiler decide that. In reality you almost never care whether out-of-bounds is available. For example the convolution filter I explained above: Superficially, it seems that this guy should require that its input provides out-if-bounds. However, it is perfectly reasonable to call it on a subinterval of an `ArrayImg` if there is enough padding around that subinterval (although `ArrayImg` provides no oob).

In reality you almost always only care whether you have access in a given interval. I don't see a way for the compiler to check that.

However, if you want to play with the idea, you could do so in ops, if @ctrueden approves.

dietzc:
> @axtimwalde   I do not think that adding the proposed method makes sense as it saves a low count of characters by cluttering the API. 

In many cases I just do the `Views.interval(Views.extend(...))` e.g. when working with neighborhoods. It's cluttering my code :-)

> @axtimwalde  Regarding imglib2 related helper methods in imagej-ops: I believe that there will be more to come and so discussing a good place in imagej-ops is probably a good idea.

I'm not sure what's a good solution. I've the feeling that it is good to avoid the situation that ImageJ-Ops serves as a simple convenience layer/package on top of `imglib2`. People shouldn't have to add a dependency on ops just the get some simple helper methods. I like `Imglib2` being self-contained. On the other hand, we will wrap `Views` (https://github.com/imagej/imagej-ops/tree/views) anyway, so we could also easily add the method there. But this is an entirely different discussion from what @LeonYang5114 wants to achieve. We could also discuss `Hyperslice` (modified version) or other basic classes to live in `ImgLib2` rather than in imagej-ops which  makes use of theses classes.

> @ctrueden  what is even the point of rebounding the interval if the random access to the outside is "lost" again? @stelfrich @dietzc @LeonYang5114

Because it's actually not really lost. See discussion and explanations of @tpietzsch above. We use this property all the time (e.g. when working with `Neighborhood`s).

> @LeonYang5114  The main reason I want this happen is that I hope we rely on the compiler to check if out-of-bound RA is allowed for a specific RAI, and fail early if not.

This won't help. The `ExtendedRAI` maybe hidden somewhere in all the `MixedTransform`, `IntervalView` etc layers of the thing you pass to your algorithm. You would have to decompose the complete object and check each source if it's an `ExtendedRAI`. Furthermore, you don't want to force any algorithm / op to work on `ExtendedRAI`. Rather you should use `Views.extend(...)` for each incoming `RAI` and let `Views` do the magic.

> @tpietzsch  However, if you want to play with the idea, you could do so in ops, if @ctrueden approves. 
I'm against diverging from `ImgLib2` regarding the basic data-structures.

LeonYang5114:
I don't understand why we would sometimes rely on the caller to provide an `ExtendedRAI` (the original one), but typed on `RandomAccessible` instead, even if we are certain that using a normal `RAI` would cause array index out of bound exception. For example in the [`Dilation`](https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/morphology/Dilation.java#L248-L250). Would it be better to just type on `ExtendedRandomAccessibleInterval` so that you know as long as the program compile, there will be no AIOOBE?

I am mentioning this because we want a `RAI` that has OOB RA, but we don't want to say it is the caller's responsibility for providing such `RAI`. Any idea?

ctrueden:
Thanks again @tpietzsch for the thorough explanation. Very helpful! I still have some questions, though:

> Now if you want to be sure to get the fastest access for a given interval you can do
> ```java
> Views.interval( accessible, interval ).randomAccess()
> ```
> and there is code that relies on it.
> With the changed policy you would need to
> ```java
> accessible.randomAccess(interval)
> ```

I'm trying to understand from the code why this is the case. The `Views.interval` methods always returns `new IntervalView` of the arguments, so even with `ExtendedRandomAccessibleInterval` extending `RandomAccessibleInterval`, there is no change at that level.

Digging into the guts of `IntervalView`, the `randomAccess()` methods lean on `TransformBuilder.getEfficientRandomAccessible`. Digging into `TransformBuilder`, there are some `isInstance` checks for things like `ExtendedRandomAccessibleInterval`, `MixedTransformView`, etc., but I don't see any instance checks on `Interval` or `RandomAccessibleInterval`.

So in short: why would this patch change the behavior as you describe above?

tpietzsch:
@ctrueden 
I think for most of what we have now, it may probably work. The point is: I'm not sure, I don't have enough spare time to invest into this, and I don't want to just take the risk of changing it. (Also, I don't see any clear benefits that would justify putting a lot time into this).

One example that I could find where you get strange results:
```
RandomAccessibleInterval< IntType > eimg = Views.extendBorder( ArrayImgs.ints( 10, 10 ) );
Cursor< IntType > c1 = Views.iterable( Views.interval( eimg, eimg ) ).cursor();
Cursor< IntType > c2 = Views.iterable( eimg ).cursor();
System.out.println( c1.getClass() ); // => net.imglib2.img.array.ArraySubIntervalCursor
System.out.println( c2.getClass() ); // => net.imglib2.view.RandomAccessibleIntervalCursor
```
(this is with `ExtendedRandomAccessibleInterval` implementing `RandomAccessibleInterval`)

axtimwalde:
@tpietzsch sorry I wasn't clear and you interpreted my question as targeted towards the `RandomAccess` requests, I think they are pretty well set as they are.  I was questioning the purpose of the suggested method in this PR to extend and again wrap the same `RandomAccessibleInterval` because requests for `RandomAccess`es on them are specified by passing along an `Interval` or not.  I see the point of convenience, but I suggest that, instead, we introduce a method `Views.iterable(RandomAccessible, Interval) and Views.flatIterable(RandomAccessible, Interval)` that clearly separates the two ideas and leaves room for all the existing optimizations.  What do you think?  It is also an addition and does not conflict with established concepts.

tpietzsch:
@axtimwalde I don't understand what you mean. What is the question, and Is your question under the assumption that we let `ExtendedRandomAccessibleInterval` implement `RandomAccessibleInterval`?

I don't see the benefit of `Views.iterable(RandomAccessible, Interval)` over `Views.iterable(Views.interval(RandomAccessible, Interval))`. It would be just a minor abbreviation, because the implementation is exactly `Views.iterable(Views.interval(RandomAccessible, Interval))`. But probably not what you meant?

I absolutely fully agree with what you said earlier, that it would be best to pass source `RandomAccessible` and `Interval` separately into methods. But with `Views.iterable()` I don't really see the point. There you really really want the source to be an Interval and no way are you ever going to access anything outside that interval. So `Views.iterable(RandomAccessibleInterval)` is the right thing to do in my opinion.

ctrueden:
@tpietzsch wrote:

> ```java
> RandomAccessibleInterval< IntType > eimg = Views.extendBorder( ArrayImgs.ints( 10, 10 ) );
> Cursor< IntType > c1 = Views.iterable( Views.interval( eimg, eimg ) ).cursor();
> Cursor< IntType > c2 = Views.iterable( eimg ).cursor();
> System.out.println( c1.getClass() ); // => net.imglib2.img.array.ArraySubIntervalCursor
> System.out.println( c2.getClass() ); // => net.imglib2.view.RandomAccessibleIntervalCursor
> ```

Because `IntervalView` is an `IterableInterval`, the above is equivalent to:

> ```java
> RandomAccessibleInterval< IntType > eimg = Views.extendBorder( ArrayImgs.ints( 10, 10 ) );
> Cursor< IntType > c1 = Views.interval( eimg, eimg ).cursor();
> Cursor< IntType > c2 = Views.iterable( eimg ).cursor();
> System.out.println( c1.getClass() ); // => net.imglib2.img.array.ArraySubIntervalCursor
> System.out.println( c2.getClass() ); // => net.imglib2.view.RandomAccessibleIntervalCursor
> ```

Sure, I agree that is a bit strange. (Thanks a lot for this example, BTW: I dug into the code & debugger quite a bit to gain a better understanding of `Views`.)

I do think this reinforces my point that no behavior would actually change, though: the call `Views.iterable( eimg )` above is of course not possible before the patch in this PR is applied. So it seems like your concern is ultimately more about what sorts of untested invocations will be come legal, versus existing invocations breaking.

> I don't have enough spare time to invest into this, and I don't want to just take the risk of changing it.

Fair enough. PR closed.

@dietzc wrote:
> We use this property all the time (e.g. when working with `Neighborhood`s).

OK, after lots of digging, I think I found what you mean, @dietzc. Please correct me if I am wrong.

In the `imglib2-algorithm` project, there is a package `net.imglib2.algorithm.neighborhood` which is what Ops uses for neighborhoods. Specifically: `MapNeighborhood` and `MapNeighborhoodWithCenter` lean on the `net.imglib2.algorithm.neighborhood.Shape` class's `neighborhoodsSafe(RandomAccessibleInterval)` method, which uses the `*NeighborhoodFactory` (e.g., `RectangleNeighborhoodFactory`) classes to create `Neighborhood` objects, based on source `RandomAccess` objects which are requested from the source based on a relevant source interval (e.g., [here is the crux for rectangle neighborhoods](https://github.com/imglib/imglib2-algorithm/blob/imglib2-algorithm-0.3.3/src/main/java/net/imglib2/algorithm/neighborhood/RectangleNeighborhoodLocalizableSampler.java#L86-L87)). The gist is that whenever there is a specified source interval, the `randomAccess(Interval)` (_not_ `randomAccess()) method is used to construct the neighborhood, which results in the out of bounds strategy being used as desired.

:tired_face:

axtimwalde:
@tpietzsch, if I get it right, the whole point of this PR is to introduce a convenience method or construct to save the 15 characters `Views.interval(` in this particular situation.  I fail to see any other reason for it.  Introducing these two methods would do that without any further side effect.  I was wrong earlier to support making `ExtendedRandomAccessibleInterval` to extend `Interval` for reasons explained by @tpietzsch. @ctrueden, thanks for closing the PR in the current form.

Issue #124, opened 2016-03-29
consider adding method for union of two intervals


bogovicj:
Intervals has a method to compute an intersection
(https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/util/Intervals.java#L206-L231)

but there is no method to compute the union. 
I'd be happy to make a pull request with this if it's of interest.

tpietzsch:
@bogovicj Yes, this is of interest and a PR would be cool!

bogovicj:
Great, will do :-)

bogovicj:
Great, thanks @tpietzsch
Fixed by pull request #125 and closes this issue.

PR #125, opened 2016-03-30
Add method for union of two intervals

bogovicj:
Re #124 
Very similar to the intersection method.
 ( "union" may not be the _best_ name, since its doesn't act like the union of sets - I'm thinking if the input is two non-overlapping intervals - but its probably clear enough )

### DIFF
--- a/src/main/java/net/imglib2/util/Intervals.java
+++ b/src/main/java/net/imglib2/util/Intervals.java
@@ -231,6 +231,33 @@ public class Intervals
 	}
 
 	/**
+	 * Compute the smallest interval that contains both input intervals.
+	 * 
+	 * Create a {@link FinalInterval} , which is the union of the input
+	 * intervals (i.e., the area contained in either input interval).
+	 * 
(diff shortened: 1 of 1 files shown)


tpietzsch:
Maybe "smallestContainingInterval" is a better name? But I'd be also okay with "union" as it's shorter and, as you said, should be clear enough. Does anyone else have an opinion on this?

In any case, please fix the javadoc:
`Create a {@link FinalInterval} , which is the union of the input intervals (i.e., the area contained in either input interval).`
This is the description of actual "union", not what the method does.

axtimwalde:
I share your initial feelings about the union, but it's used in the same spirit in https://docs.oracle.com/javase/7/docs/api/java/awt/Rectangle.html#union%28java.awt.Rectangle%29 therefore and in the interest of short names I am strongly pro union ;).

bogovicj:
Then we'll stick with "union"
Javadoc edited!

tpietzsch:
thanks @bogovicj

PR #128, opened 2016-04-22
added RandomAccessiblePair< A, B >, a trivial pairing of two sources into a source of pairs

axtimwalde:
Enables convenient composition of arbitrary types which I found incredibly
useful, e.g. converting a FloatType channel and a ByteType channel into an
ARGBType channel using Converter<Pair<FloatType, ByteType>, ARGBType>.

### DIFF
--- a/src/main/java/net/imglib2/view/RandomAccessiblePair.java
+++ b/src/main/java/net/imglib2/view/RandomAccessiblePair.java
@@ -1,0 +1,247 @@
+package net.imglib2.view;
+
+import net.imglib2.Interval;
+import net.imglib2.Localizable;
+import net.imglib2.RandomAccessible;
+import net.imglib2.util.Pair;
+
+/**
+ * A {@link RandomAccessible} over two independent
+ * {@link RandomAccessible RandomAccessibles} whose type is the {@link Pair} of
+ * corresponding values at the same coordinates in either of the two sources.
+ *
+ * @author Stephan Saalfeld &lt;saalfelds@janelia.hhmi.org&gt;
+ */
+public class RandomAccessiblePair< A, B > implements RandomAccessible< Pair< A, B > >
+{
+	public class Pair implements net.imglib2.util.Pair< A, B >
+	{
+		private A a = null;
+		private B b = null;
+
+		@Override
+		public A getA()
+		{
+			return a;
+		}
+
+		@Override
+		public B getB()
+		{
+			return b;
+		}
+
+		public void setA( final A a )
+		{
+			this.a = a;
+		}
+
+		public void setB( final B b )
+		{
+			this.b = b;
+		}
+	}
+
+	public class RandomAccess implements net.imglib2.RandomAccess< net.imglib2.util.Pair< A, B > >
+	{
+		final protected net.imglib2.RandomAccess< A > raA;
+		final protected net.imglib2.RandomAccess< B > raB;
+		final protected Pair pair;
+
+		public RandomAccess()
+		{
+			raA = sourceA.randomAccess();
+			raB = sourceB.randomAccess();
+			pair = new Pair();
+		}
+
+		@Override
+		public void localize( final int[] position )
+		{
+			raA.localize( position );
+		}
+
+		@Override
+		public void localize( final long[] position )
+		{
+			raA.localize( position );
+		}
+
+		@Override
+		public int getIntPosition( final int d )
+		{
+			return raA.getIntPosition( d );
+		}
+
+		@Override
+		public long getLongPosition( final int d )
+		{
+			return raA.getLongPosition( d );
+		}
+
+		@Override
+		public void localize( final float[] position )
+		{
+			raA.localize( position );
+		}
+
+		@Override
+		public void localize( final double[] position )
+		{
+			raA.localize( position );
+		}
+
+		@Override
+		public float getFloatPosition( final int d )
+		{
+			return raA.getFloatPosition( d );
+		}
+
+		@Override
+		public double getDoublePosition( final int d )
+		{
+			return raA.getDoublePosition( d );
+		}
+
+		@Override
+		public int numDimensions()
+		{
+			return RandomAccessiblePair.this.numDimensions();
+		}
+
+		@Override
+		public void fwd( final int d )
+		{
+			raA.fwd( d );
+			raB.fwd( d );
+		}
+
+		@Override
+		public void bck( final int d )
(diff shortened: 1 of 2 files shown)


src/main/java/net/imglib2/view/RandomAccessiblePair.java
```diff
@@ -0,0 +1,247 @@
...
+/**
+ * A {@link RandomAccessible} over two independent
+ * {@link RandomAccessible RandomAccessibles} whose type is the {@link Pair} of
+ * corresponding values at the same coordinates in either of the two sources.
+ *
+ * @author Stephan Saalfeld &lt;saalfelds@janelia.hhmi.org&gt;
+ */
+public class RandomAccessiblePair< A, B > implements RandomAccessible< Pair< A, B > >
+{
+	public class Pair implements net.imglib2.util.Pair< A, B >
```
tpietzsch: just use Pair instead of net.imglib2.util.Pair

src/main/java/net/imglib2/view/RandomAccessiblePair.java
```diff
@@ -0,0 +1,247 @@
...
+		}
+
+		public void setB( final B b )
+		{
+			this.b = b;
+		}
+	}
+
+	public class RandomAccess implements net.imglib2.RandomAccess< net.imglib2.util.Pair< A, B > >
+	{
```
tpietzsch: same here

src/main/java/net/imglib2/view/Views.java
```diff
@@ -1234,4 +1235,19 @@ public static boolean isZeroMin( final Interval interval )
...
 
+	/**
+	 * Compose two {@link RandomAccessible} sources into a
+	 * {@link RandomAccessible} of {@link Pair}.
+	 *
+	 * @param sourceA
+	 * @param sourceB
+	 * @return
+	 */
+	public static < A, B > RandomAccessiblePair< A, B > pair(
```
tpietzsch: I would suggest to make the return type `RandomAccessible<Pair<A, B>>` or even `RandomAccessible<? extends Pair<A, B>>`, if there is no strong need for being _exactly_ `RandomAccessiblePair<A, B>`. I can imagine that we might want to add other variants implementing this later.

src/main/java/net/imglib2/view/RandomAccessiblePair.java
tpietzsch: I'm not sure I like the name overloading

tpietzsch:
I knew that sounded very familiar... I did the same thing a long time ago on a branch...
https://github.com/imglib/imglib2/blob/roi-experiments/roi/src/main/java/net/imglib2/newroi/util/PairRandomAccess.java
https://github.com/imglib/imglib2/blob/roi-experiments/roi/src/main/java/net/imglib2/newroi/util/PairRandomAccessible.java
So, I guess I like the idea :-)

I my variant, I simply made the `PairRandomAccess` implement `Pair`, so there is no need for a separate `Pair` object. I like that better, because it postpone's access to `get()` on the source accesses until you actually query the Pair.

StephanPreibisch:
In general a good thing to have, one way or the other

src/main/java/net/imglib2/view/Views.java
axtimwalde: Thanks.  Fixed with making the `RandomAccess` implement `Pair` as suggested below.

src/main/java/net/imglib2/view/RandomAccessiblePair.java
axtimwalde: Pair removed as suggested below.

axtimwalde:
Thanks @tpietzsch for your comments.  Yes, this is not the first time I implemented that either ;).  The suggestion to make the RandomAccess be a Pair is very nice.  I changed it so and added you as author.

tpietzsch:
@axtimwalde thanks!

tpietzsch:
@axtimwalde this is really nice!
To copy img to img2 I can now
```
Views.interval(Views.pair(img, img2), img).forEach(t -> t.getB().set(t.getA()));
```
The `Views.interval` on top is a bit cumbersome.

I wouldn't add a variant `RAI Views.pair(RAI, RAI)` because that needs checking compatibility of the input intervals, throwing exception if not, etc, is not really cool.

Instead: a `Views.pair` variant `RAI Views.pair(RAI, RA)` that imposes interval of first argument seems useful. It couldn't hurt IMO to have the result be iterable too (because it always can be). Prototype with existing classes would be
```
	public static < A, B > IntervalView< Pair< A, B > > pair(
			final RandomAccessibleInterval< A > sourceA,
			final RandomAccessible< B > sourceB )
	{
		return Views.interval( new RandomAccessiblePair< A, B >( sourceA, sourceB ), sourceA );
	}
```
And voilà:
```
	Views.pair(img, img2).forEach(t -> t.getB().set(t.getA()));
```
What do you think? Should we add this? @axtimwalde @StephanPreibisch @dietzc @ctrueden

tpietzsch:
BTW, I think this method of copying would be quite a bit slower than doing it with two cursors and a loop. Still cool that one can write it like this... :-)

dietzc:
@tpietzsch that is actually really cool. We can reuse this stuff in the `Map` implementations of `Op`s.

axtimwalde:
@tpietzsch, how about
```java
public static < A, B > IntervalView< Pair< A, B > > pair(
            final RandomAccessible< A > sourceA,
            final RandomAccessible< B > sourceB,
            final Interval interval )
    {
        return Views.interval( new RandomAccessiblePair< A, B >( sourceA, sourceB ), interval );
    }
```
I would like this better because it is more flexible (I can pass `sourceA` or `sourceB` or another `Interval` depending on who is what and it has almost no extra-characters in useage.  It is actually less characters in cases where explicit casting is required to convince the compiler to pick the right method signature... (that's another story that bugged me recently https://github.com/saalfeldlab/bigcat/blob/janh5/src/main/java/bdv/bigcat/ui/ARGBConvertedLabelPairSource.java#L37)

tpietzsch:
@axtimwalde Yes, the converters ambiguity I also ran into... But it's `IterableInterval` vs `RandomAccessibleInterval`. In the Views case it would work though, for `RandomAccessibleInterval` and `RandomAccessible` it will just pick the most specific one.

However, I'm also ok with the explicit interval parameter. (Or with not adding a new method at all.)

PR #132, opened 2016-04-25
added ValueEquals interface to test value equivalency

axtimwalde:
added ValueEquals interface to test value equivalency,
made Type extend this interface and implemented it across all core types,
bumped version to 3.0.0-SNAPSHOT

### DIFF
--- a/src/main/java/net/imglib2/view/composite/AbstractNumericComposite.java
+++ b/src/main/java/net/imglib2/view/composite/AbstractNumericComposite.java
@@ -193,4 +193,21 @@ abstract public class AbstractNumericComposite< T extends NumericType< T >, C ex
 		sourceAccess.setPosition( -1, d );
 		return iterator;
 	}
+
+	@Override
+	public boolean valueEquals( C t )
+	{
+		if ( t.length == length )
+		{
+			final Iterator< T > a = iterator();
+			final Iterator< T > b = iterator();
+			while ( a.hasNext() )
+				if ( !a.next().valueEquals( b.next() ) )
+					return false;
+
+			return true;
+		}
+		else
+			return false;
+	}
 }
--- a/src/main/java/net/imglib2/view/composite/AbstractNumericComposite.java
+++ b/src/main/java/net/imglib2/view/composite/AbstractNumericComposite.java
@@ -200,7 +200,7 @@ abstract public class AbstractNumericComposite< T extends NumericType< T >, C ex
 		if ( t.length == length )
 		{
 			final Iterator< T > a = iterator();
-			final Iterator< T > b = iterator();
+			final Iterator< T > b = t.iterator();
 			while ( a.hasNext() )
 				if ( !a.next().valueEquals( b.next() ) )
 					return false;
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
 	</parent>
 
 	<artifactId>imglib2</artifactId>
-	<version>2.9.1-SNAPSHOT</version>
+	<version>3.0.0-SNAPSHOT</version>
 
 	<name>ImgLib2 Core Library</name>
 	<description>A multidimensional, type-agnostic image processing library.</description>
--- a/src/main/java/net/imglib2/type/Type.java
+++ b/src/main/java/net/imglib2/type/Type.java
@@ -34,6 +34,8 @@
 
 package net.imglib2.type;
 
+import net.imglib2.type.operators.ValueEquals;
+
 /**
  * The {@link Type} interface is responsible for accessing and manipulating
  * (pixel) values. It can be realized as an object instance holding a single
@@ -47,7 +49,7 @@ package net.imglib2.type;
  * @author Stephan Preibisch
  * @author Stephan Saalfeld
  */
-public interface Type< T extends Type< T > >
+public interface Type< T extends Type< T > > extends ValueEquals< T >
 {
 	/**
 	 * Creates a new {@link Type} variable which can only store one value.
--- a/src/main/java/net/imglib2/type/label/BasePairBitType.java
+++ b/src/main/java/net/imglib2/type/label/BasePairBitType.java
@@ -39,6 +39,7 @@ import net.imglib2.img.NativeImgFactory;
 import net.imglib2.img.basictypeaccess.LongAccess;
 import net.imglib2.type.AbstractBit64Type;
 import net.imglib2.type.BasePairType;
+import net.imglib2.type.logic.BitType;
 
 /**
  * Representation of base pairs using 3 bits per entry, supported characters: gap, N, A, T, G, C, U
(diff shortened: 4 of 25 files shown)


axtimwalde:
Note that this change requires to raise the major version.  It is possible ti increase only the minor version by not having `Type extends ValueEquals` and have only each leaf implementation of `Type` implement this interface individually.  Implementations at the abstract class level would again require to raise the major version. I do not like this option and would prefer the major version increase.

ctrueden:
What if you add a `default` method to the interface which simply returns false?

Before Java 8, our policy at the ImageJ2 layer was that if we added a new method to an interface, but added a default implementation to the corresponding abstract class(es), then we did not bump the major version.

There is a difference between __API__ compatibility—whether downstream _calling code_ would break—and __SPI__ compatibility—whether downstream _extending code_ would break. We opt to have SemVer mean __API__ compatibility, although we make a best effort for SPI compatibility as well as described above.

axtimwalde:
@ctrueden, that would make imglib2 depend on Java8 for which there is currently no reason.  Regardless of this, I do not like the default implementation here for that it hides undesired behavior from unaware consumers which is generally a bad thing.

axtimwalde:
ImgLib2's 'API' being merely interfaces I also strongly vote for SPI compatibility.

ctrueden:
> that would make imglib2 depend on Java8 for which there is currently no reason.

Sorry, I was going on the assumption that ImgLib2 would switch to Java 8 imminently, as indicated by @tpietzsch in chat earlier today. Apologies if I misunderstood him.

> ImgLib2's 'API' being merely interfaces I also strongly vote for SPI compatibility.

Sure. I think we should document this on a per-component (or per-org) basis here somewhere.

axtimwalde:
@ctrueden thanks for clarification.  I am not aware of an ImgLib2 -> Java8 discussion in a chat today.  Will wait for @tpietzsch's comments on this.

tpietzsch:
@ctrueden @axtimwalde The imglib2 and imglib2-algorithm artefacts I released today are java8. I saw no reason to delay this any longer. I see no point of sticking with java6. We shouldn't deny ourselves the new java8 features and support a java version that is itself not supported anymore.

Regarding the original discussion: I didn't look at the PR yet. I'll comment tomorrow.

axtimwalde:
@tpietzsch I agree in principle to go for Java8 whenever needed.  Is there Java8 code in imglib2 and imglib2-algorithms?  If not, we unnecessarily prevent them to be used in pre Java 8 environments.  Such an environment could be Mac OS or Ubuntu 15.04 LTS.

axtimwalde:
@ctrueden interface heavy, i.e. interfaces make a big part of the API.  The second argument that the suggested default would generate undesired surprises is stronger.  Let's see what @tpietzsch has to say.

tpietzsch:
@axtimwalde There is no java8 code, yet. However, there might be in the future. There will never really be a strong *need* to go Java8 because for every individual feature we want to have we can work around it with java6. I mean stuff like the need to have both `copy()` and `copyCursor()`. At least that holds for the core. That could in principle be java6 forever.

For other imglib2 components it will be more difficult. For components that use 3rd party libraries, it will mean that we cannot use the most recent versions of those libraries. This will eventually lead to clashes with other things in Fiji which require the more recent version, and then we or someone on the Fiji side will need to work around _that_.

This is about developer convenience vs user convenience. It is certainly possible to install Java 8 on Ubuntu 14.04 LTS and older, it's just more inconvenient for the user. At some point developer convenience outweighs user convenience. One can argue whether we are at that point already, but I think we should move on now.

src/main/java/net/imglib2/view/composite/AbstractNumericComposite.java
```diff
@@ -193,4 +193,21 @@ public void mul( final double c )
...
 		return iterator;
 	}
+
+	@Override
+	public boolean valueEquals( C t )
+	{
+		if ( t.length == length )
+		{
+			final Iterator< T > a = iterator();
+			final Iterator< T > b = iterator();
```
tpietzsch: One of these should be `t.iterator()`.

tpietzsch:
Phew. This is a tough one.

I'm not sure whether ValueEquals should be a superclass of Type. I would tend to either keep it separate, or add the valueEquals method directly to the type interface.

I'm on the fence about the default method. On the one hand, it would be nice to have, because it wouldn't break downstream code. On the other hand, we might _actually want to break downstream code_ to make implementors of `Type` think about what to do. Actually I tend to doing the later thing. Then maybe a few releases down the line we can add the default method anyway. I would not have it return false but call `equals`.

About the version, I honestly don't care. Why not bump to 3.0.0? Who cares...

src/main/java/net/imglib2/view/composite/AbstractNumericComposite.java
axtimwalde: Yes, fixing it.

axtimwalde:
Ok, version topic is settled, bump to 3.0.0 if we make it superclass of `Type`.

Regarding the superclass issue.  I was thinking about it for some time and could not find an example to have any `Type` not implement this interface but lots of examples where `ValueEquals` would not be a `Type`.  I think keeping it a separate interface would also be even less strict.  If my observation is true and all `Type` should implement `ValueEquals` superclassing `Type` is more convenient (saving some & in generics).  I will make a version that separates it from `Type` and see how that looks in comparison.

tpietzsch:
@axtimwalde don't bother. Let's leave it like this if you think that `ValueEquals` is useful outside `Type`, or put the method directly into type is it's not.

axtimwalde:
@tpietzsch, just started looking at it.  If I were to not superclass `Type`, I do not know where to start superclassing.  The next level is `BasePairType`, `BooleanType`, `NativeType`, `NumericType`, all good candidates but with the same consequence of introducing something new into interfaces.  The next level seems even more arbitrary, so I would go for implementing it only in leaves which I do not like for how messy and unstructured it is.  So I go for the 'don't bother', fix the bug and stick with how it is now.  `ValueType` as a separate interface makes sense e.g. for `Neighborhood` or other `IterableInterval`s.

tpietzsch:
@axtimwalde ok, great. `Neighborhood` makes sense, yes. Good example

ctrueden:
> we unnecessarily prevent them to be used in pre Java 8 environments. Such an environment could be Mac OS or Ubuntu 15.04 LTS.

Well, Ubuntu can run Java 8 since at least 12.04 LTS. We ship Java 8 with Fiji and AFAIK it runs fine on older Ubuntus. The main problem is OS X 10.7 Lion and earlier. And Windows 2000 maybe? 💻 

https://docs.google.com/spreadsheets/d/19rSofTznt8XT6hRUPKmDspayUYdSQujfTjVm1NqYQP4/edit?usp=sharing

PR #135, opened 2016-07-11
ARGBChannelSampleConverters to make single channels or sets of channels of an ARGBType image available as UnsignedByteType dimensions

axtimwalde:


axtimwalde:
@tpietzsch OK with me merging this?

tpietzsch:
@axtimwalde Yes, cool that the SamplerConverters get used finally!

PR #136, opened 2016-07-14
implemented equals() and hashCode() in ValuePair 

hoerldavid:
Pairs should be equal if their contents are equal. 
The proposed methods are slightly modified from Eclipses auto-generated methods so that it will equal any Pair with the same contents.


tpietzsch:
thanks David!

Issue #141, opened 2016-08-29
More helpful Views exceptions when inputs are invalid


etarena:
We invoking Views methods it would be nice to have more input validation upfront to avoid exceptions like the following:
```
java.lang.ArrayIndexOutOfBoundsException: 2
	at net.imglib2.AbstractInterval.min(AbstractInterval.java:182)
	at net.imglib2.util.Intervals.contains(Intervals.java:374)
	at net.imglib2.img.array.ArrayImg.fastCursorAvailable(ArrayImg.java:188)
	at net.imglib2.img.array.ArrayImg.supportsOptimizedCursor(ArrayImg.java:230)
	at net.imglib2.view.iteration.IterableTransformBuilder.buildIterableInterval(IterableTransformBuilder.java:237)
	at net.imglib2.view.iteration.IterableTransformBuilder.getEfficientIterableInterval(IterableTransformBuilder.java:77)
	at net.imglib2.view.IntervalView.getFullViewIterableInterval(IntervalView.java:137)
	at net.imglib2.view.IntervalView.iterationOrder(IntervalView.java:156)
	at net.imglib2.view.Views.flatIterable(Views.java:765)
	at net.imglib2.img.ImgView.<init>(ImgView.java:81)
```
This exception results from an attempt to call `Views.interval(img, min, max)` where `img.numDimensions() == 3` and `min.length == max.length == 2`.

axtimwalde:
Please test these conditions outside of the Views framework where necessary.  We do not want to perform runtime tests where inputs can be guaranteed to be correct.

PR #142, opened 2016-08-29
Views dimension validation

etarena:
This branch addresses #141, adding dimensionality checks for Views methods.

### DIFF
--- a/src/main/java/net/imglib2/view/IntervalView.java
+++ b/src/main/java/net/imglib2/view/IntervalView.java
@@ -11,13 +11,13 @@
  * %%
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
- * 
+ *
  * 1. Redistributions of source code must retain the above copyright notice,
  *    this list of conditions and the following disclaimer.
  * 2. Redistributions in binary form must reproduce the above copyright notice,
  *    this list of conditions and the following disclaimer in the documentation
  *    and/or other materials provided with the distribution.
- * 
+ *
  * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
  * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
  * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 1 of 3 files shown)


axtimwalde:
Hi @etarena , we chose to not test various conditions at runtime to make Views as performant as possible.  Assertions are in place to test the minimal requirements for debugging.  I am therefore voting strongly against accepting this pull request.  If you need to perform these tests at runtime in an application, please do so outside of imglib2 core.

ctrueden:
@axtimwalde Do you have any real use cases where you need to call Views millions of times where these checks will actually make a significant difference?

As things stand, Views is _very_ heavily used and recommended, and yet when you make simple common mistakes such as passing a 2D min/max array with a 3D image, you get a totally arcane and impenetrable exception. This sucks, and we can do better.

axtimwalde:
@ctrueden yes, we are calling views rapidly during optimization.  The tests do not make a big difference, I agree, but we generally do not perform runtime input validations in ImgLib2 anywhere to maximize algorithm performance.  Helpful debugging output is supported via assertions.  I suggest to perform runtime tests outside of core closer to applications where this is desired, e.g. Ops.  Does that work for you?

ctrueden:
> I suggest to perform runtime tests outside of core closer to applications where this is desired, e.g. Ops. Does that work for you?

OK.

StephanPreibisch:
I agree with @axtimwalde ... once we start checking things it is unclear where to draw the line of what to check, which at the end makes the framework more unpredictable in my opinion, even not considering performance aspects.

tpietzsch:
@StephanPreibisch That was exactly my concern with this PR. I was actually tempted to just merge it because in this case it wouldn't be much overhead, but it is difficult to know where to stop. It's cleaner to have a policy of not doing safety checks anywhere.

ctrueden:
I am willing to respect this decision in the ImgLib2 space. But I do feel compelled to chime in that I disagree with this sort of [slippery slope](https://en.wikipedia.org/wiki/Slippery_slope) argument. This is a gray area of balancing performance with usability. My argument here, and the reason for this PR, is because @etarena and I found `Views` to be difficult to use due to the arcane `ArrayIndexOutOfBoundsException`, and thought that adding some basic dimensional length checks would make things much easier for users, especially new users such as @etarena, while not impacting performance negatively much if at all. This was not the first time I have received, and/or seen others receive, these sorts of arcane exceptions. They contribute to the perception of new users that ImgLib2 is deep and complicated and hard to learn and use—which I think has a lot of truth actually, and I would like us to do whatever we can to make it easier to adopt and use.

On my end, I will do as @axtimwalde suggests and shift that balance more toward usability in ImageJ, especially ImageJ Ops. Then I can simply tell new users to use ImageJ, since it builds a lot of niceties on top of ImgLib2 and is much easier to learn.

PR #143, opened 2016-09-13
Add LinearAccessible interface and Views

hanslovsky:
`LinearAccessible` can be accessed through a linear index by a `LinearAccess`. This allows very fast access for certain types of `RandomAccessibleInerval`, e.g. `ArrayImg`. `Views.linearAccess` allows to access any `RandomAccessibleinerval` by linear index, using fast implementations where available.

### DIFF
--- a/src/main/java/net/imglib2/LinearAccess.java
+++ b/src/main/java/net/imglib2/LinearAccess.java
@@ -1,0 +1,93 @@
+/*
+ * #%L
+ * ImgLib2: a general-purpose, multidimensional image processing library.
+ * %%
+ * Copyright (C) 2009 - 2016 Tobias Pietzsch, Stephan Preibisch, Stephan Saalfeld,
+ * John Bogovic, Albert Cardona, Barry DeZonia, Christian Dietz, Jan Funke,
+ * Aivar Grislis, Jonathan Hale, Grant Harris, Stefan Helfrich, Mark Hiner,
+ * Martin Horn, Steffen Jaensch, Lee Kamentsky, Larry Lindsey, Melissa Linkert,
+ * Mark Longair, Brian Northan, Nick Perry, Curtis Rueden, Johannes Schindelin,
+ * Jean-Yves Tinevez and Michael Zinsmaier.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ * #L%
+ */
+package net.imglib2;
+
+import net.imglib2.img.array.ArrayImg;
+
+/**
+ *
+ * @author Philipp Hanslovsky
+ *
+ *         Access a source {@link LinearAccessible} by linear indexing. An
+ *         example use case is array like access for an {@link ArrayImg}.
+ *
+ * @param <T>
+ */
+public interface LinearAccess< T > extends Sampler< T >
+{
+
+	@Override
+	LinearAccess< T > copy();
+
+	/**
+	 *
+	 * Set position at which data be read on call to {@link Sampler#get()}.
+	 */
+	void set( int position );
+
+	/**
+	 *
+	 * Set position at which data be read on call to {@link Sampler#get()}.
+	 */
+	void set( long position );
+
+	/**
+	 *
+	 * Set position and immediately return {@link T} at this position.
+	 */
+	default T get( final int position )
+	{
+		set( position );
+		return get();
+	}
+
+	/**
+	 *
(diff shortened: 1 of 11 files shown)


axtimwalde:
Thanks @hanslovskyp, not yet merge ready, so I am closing this pull request.
What do you think about not introducing an additional interface but instead implementing the exact same logic in a View that converts an n-dimensional RAInterval into a 1-dimensional RAInterval.  The RAccess on this would be what you implemented here, accepting 0/ ignoring the dimension parameter in setPosition, move and fwd, bck methods?  Also, please do not add curly brackets to single line loops.

axtimwalde:
So it is mostly renaming, removing the LinearAccessible and LinearAccess interfaces and move the logic to RandomAccessibleInterval and RandomAccess implementations.

hanslovsky:
I agree that keeping the number of interfaces as low as possible is desirable. However, having a (useless) dimension parameter for `setPosition` would be confusing. With good documentation the usage should be clear. I prefer just ignoring the dimension parameter for performance reasons.
What about `setPosition(long[])` and similar methods? Throw `UnsupportedOperationException`?
A good name for the `Views` method would then be `Views.flatten`.

hanslovsky:
One thing that I do not like about using RAI instead of LinearAccessible is that method signatures will not tell the caller whether they expect an actual RAI or a flattened RAI.

tpietzsch:
General question: Where is this useful where it cannot be replaced by (flat) iteration easily?

hanslovsky:
I am working on a n-d watershed and for performance reasons (probably only true for `ArrayImg`) I would like to access a RAI in a linear fashion like an array.

axtimwalde:
@tpietzsch iterables do not offer efficient modes of random access into a linearized index, being able to enumerate and access an integer coordinate by a single index can be useful, @hanslovskyp needs it for an efficient compact watershed implementation.  I do, though prefer the idea of not introducing a new interface but instead doing the same thing with RandomAccess, we can do that first outside ImgLib2 and if it turns out to be generic enough, think again about a pull request.

tpietzsch:
What about adding index access to `Cursor`.

"`setPosition`" could be trivially implemented for existing cursors, even be added to the `Cursor` interface as a default method. Plus optimized implementations for `ArrayCursor` etc.

"`getPosition`" is not as trivial, but I'm sure it could be done without or with minimal overhead.

tpietzsch:
and by the way: awesome that you are working on watershed! parallel?

hanslovsky:
`getPosition` is not an immediate requirement and could be dropped. So we could just extend `Cursor` with a `setPosition` method. Extending an existing interface is a potentially big change. Do you think that (currently) single use case justifies this extension?

Unfortunately, no parallel watershed so far, but @axtimwalde and I certainly have that in mind and have discussed possible parallel solutions. So far we have not come up with a method that would be parallel and guarantee the same solution as a single threaded watershed on the whole volume, unfortunately.

tpietzsch:
> getPosition is not an immediate requirement and could be dropped. So we could just extend Cursor with a setPosition method. Extending an existing interface is a potentially big change. Do you think that (currently) single use case justifies this extension?

I'm not sure. @axtimwalde @StephanPreibisch @dietzc @ctrueden what do you think? Would we have other use cases?

> Unfortunately, no parallel watershed so far, but @axtimwalde and I certainly have that in mind and have discussed possible parallel solutions. So far we have not come up with a method that would be parallel and guarantee the same solution as a single threaded watershed on the whole volume, unfortunately.

Yes, it's not straightforward. I was also thinking about it some time ago. It is related to component trees (obviously) and there are several papers about parallelizing that. All some variations of "do-something-stupid-locally-then-union-find-to-merge". It's no longer linear than but I could imagine it could even be an improvement in practice when running single-threaded, because you don't have to maintain large queues and access all over the place all the time.

dietzc:
Ok, maybe it's a super stupid and naive question (caused by jetlag or whatever): but basically you want to have a first version of `reset()` + `jmpForward(int steps)` methods which are provided in the `Iterator` interface? why not add `setIndex(long idx)` to this interface which in the default implementation does `reset()` and `jumpFwd(...)` and in the special case (like `ArrayImg`) does it in a more efficient way?

PS: Concerning the watershed... maybe you can consolidate efforts with https://github.com/imagej/imagej-ops/pull/446.

axtimwalde:
@tpietzsch we are currently interested in extensions of the watershed cost functions that make parallelization even harder, in particular two variants of 'compact watershed' where the Euclidean or geodesic distance to the respective seeds contributes to the cost.  For the 'classic' watershed, I believe that this approach

http://www.sciencedirect.com/science/article/pii/S0031320399001545

would enable straight forward parallelization?  Haven't tried yet.

We could add index access to cursor, but I am not yet sure that this is actually better than unfolding dimensions and sticking with `RandomAccess`.  I actually found myself wishing that I could unfold the third dimension into the second recently (to save volume blocks as JPG).  The 1-dimensional RandomAccess is the terminal case for this kind of flattening and we could make it available just so.  I like a `RandomAccess` solution for that it is consistent with the current thinking of random access vs. iteration.  I like the `Cursor` solution for that it does not ask you to pass arrays or dimension specifiers when you already know that there is only 1 dimension.

tpietzsch:
@dietzc yes, that was exactly what I was thinking too!

tpietzsch:
@axtimwalde a difference of `Cursor` solution vs `RandomAccess` is that `Cursor` indices are well-defined wrt the min of the interval. For min=(0,0,...) it's always easy, but to it is (to me) not immediately clear how to flatten dimensions for non-0-min intervals.

tpietzsch:
> I believe that this approach
> 
> http://www.sciencedirect.com/science/article/pii/S0031320399001545
> 
> would enable straight forward parallelization? Haven't tried yet.

I don't know that paper. I'll have a look tomorrow!
Can you point me to papers regarding the " 'compact watershed' where the Euclidean or geodesic distance to the respective seeds contributes to the cost"?

hanslovsky:
This is the original reference:
https://www.tu-chemnitz.de/etit/proaut/forschung/rsrc/cws_pSLIC_ICPR.pdf

Essentially, you remember the position of the seeds (or the number of steps you had to go so far to reach that pixel) and add that distance to the gray value difference.

ctrueden:
> @ctrueden what do you think?

I just wonder why we need to have both `randomAccess()` and `cursor()` if they are both converging on the same API in the end?

tpietzsch:
@ctrueden as @dietzc pointed out, you can already always "fake" `c.setPosition(n)` by `c.reset(); c.jumpFwd(n+1)`. The question is whether we want to add a (possibly more efficient) shortcut for that.

ctrueden:
> you can already always "fake"

OK, right.

> The question is whether we want to add a (possibly more efficient) shortcut for that.

Makes sense to me!

axtimwalde:
@tpietzsch @dietzc: @hanslovsky and me have discussed this again and we currently think that the `flatten`-ed accessible should be a `RandomAccessibleInterval` with `min=[0]`.  Reasons are:
* no change to the core API of ImgLib2,
* no inefficient default implementation that will depress developers and is aesthetically  unpleasing (see below),
* the index in `Cursor` is not a coordinate like in `RandomAccess`, i.e. localize on a `Cursor` is not expected to return the index but an n-dimensional location,
* `Cursor.reset()` sets the `Cursor` to index -1 instead 0 for making it compatible with Java's `Iterator` logic (`next()`, `hasNext()`), i.e. the default implementation for `setIndex(idx)` is `cursor.reset(); jumpFwd(++idx);`,
* the 0-coordinate of the `flatten`-ed `RandomAccessibleInterval` will map to the min corner of the source `RandomAccessibleInterval`, simple contract,
* we can extend the concept of flattening all dimensions to flattening one dimension into another

dietzc:
> no change to the core API of ImgLib2. 

We could use Java8 default methods in interfaces to avoid any API breaks to implement default behaviour. The nice thing is that adding `setIndex(...)` in the `Iterator` make it easy for implementing classes like `ArrayImg` to offer an efficient implementation. If you would do it in `Views.flatten(...)` you would have to special case on a certain interface which you have to introduce or for each implementation on it's own to get efficient implementations. Maybe I'm missing something here.

Anyway, I totally see the use-cases now :+1:

PR #149, opened 2017-02-25
Add expand method to Intervals 

dietzc:
... which expands an `Interval` in in all dimensions given `Dimensions` object.

### DIFF
--- a/src/main/java/net/imglib2/util/Intervals.java
+++ b/src/main/java/net/imglib2/util/Intervals.java
@@ -150,6 +150,33 @@ public class Intervals
 		}
 		return new FinalInterval( min, max );
 	}
+	
+	/**
+	 * Grow/shrink an interval in all dimensions.
+	 * 
+	 * Create a {@link FinalInterval}, which is the input interval plus border
+	 * pixels on every side, in every dimension.
+	 * 
+	 * @param interval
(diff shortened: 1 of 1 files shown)


hanslovsky:
Would it make sense to expand min and max differently?

tpietzsch:
@hanslovsky I agree that vararg overload would be useful. Could you add this, @dietzc?

Expanding min and max differently:  I wouldn't do that. In that case you would pass an `Interval`, and use the distance of min/max to the origin to determine where you want to add what. But this would allow also to shrink and/or shift the interval. So `expand` isn't the right name for that. It might lead to confusion if you pass an `Interval` and expect it to be treated as a `Dimensions`.

dietzc:
of course. like this?

tpietzsch:
👍

hanslovsky:
I just noticed, unfortunately after the PR was already merged: The added methods are not static.

dietzc:
I'll fix it. Give me 1h.

dietzc:
https://github.com/imglib/imglib2/commit/027f5090c2c493f58b584c2b3d2df8f8c29add6a

hanslovsky:
That hour went by fast. :+1: Looks good to me!

PR #151, opened 2017-03-13
Bugfix: copied CompositeRandomAccess was not positioned correctly

tpietzsch:


tpietzsch:
see https://github.com/imglib/imglib2-algorithm/issues/34

PR #152, opened 2017-03-13
Cellrevision

tpietzsch:
Revised CellImg. This is related to revision/transition from BigDataViewer cache to imglib2-cache. Some abstract classes are not needed anymore. Cell coordinate math has been separated out of the main CellImg. And so on.

Also fixed javadoc errors for java 8


ctrueden:
Hooray for progress! Too bad about the WIPs though!

PR #153, opened 2017-03-20
Add convenience functions for Accesses to ArrayImgs

hanslovsky:
This commit adds method signatures like
```java
< A extends IntAccess > void ArrayImgs.ints( A access, final long... dim );
```
This makes using accesses other than `IntArray` etc. more convenient and would be useful for the accesses implemented in `imglib2-unsafe` and for usage in `imglyb` (c-python imglib).

### DIFF
--- a/src/main/java/net/imglib2/img/array/ArrayImgs.java
+++ b/src/main/java/net/imglib2/img/array/ArrayImgs.java
@@ -34,6 +34,12 @@
 
 package net.imglib2.img.array;
 
+import net.imglib2.img.basictypeaccess.ByteAccess;
+import net.imglib2.img.basictypeaccess.DoubleAccess;
+import net.imglib2.img.basictypeaccess.FloatAccess;
+import net.imglib2.img.basictypeaccess.IntAccess;
+import net.imglib2.img.basictypeaccess.LongAccess;
+import net.imglib2.img.basictypeaccess.ShortAccess;
 import net.imglib2.img.basictypeaccess.array.ByteArray;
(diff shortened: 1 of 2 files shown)


tpietzsch:
Great, thanks. This will be also useful for the CacheLoader stuff

ctrueden:
This is cool. But, should the old methods now delegate to the new ones? E.g.:
```java
/**
 * Creates an {@link ArrayImg}&lt;{@link UnsignedByteType}, {@link ByteArray}&gt;
 * reusing a passed byte[] array.
 */
final public static ArrayImg< UnsignedByteType, ByteArray > unsignedBytes( final byte[] array, final long... dim )
{
	return unsignedBytes( new ByteArray( array ), dim );
}
```
This would reduce code duplication and be easier to read at a glance.

hanslovsky:
@ctrueden I thought about that.

In the end I decided against it because I tested against the existing methods as a sanity check. Those tests would become obsolete.

Readability is a valid point, though, and I'd be willing to update the tests. More opinions?

ctrueden:
> In the end I decided against it because I tested against the existing methods as a sanity check. Those tests would become obsolete.

I think we still want those tests. The point of unit tests is that they are agnostic of the implementation. We keep them, in case behavior changes again in the future. Guards against regressions.

> Readability is a valid point, though

Also correctness, consistency and ease of refactoring later. If the code needs to change on how the access objects get wrapped into `Img`s, we would then only need to change the code in one place (per type), instead of two.

PR #154, opened 2017-03-20
IterablePair: add a pairwise iteration utility class

etarena:
This helper class iterates over two iterations simultaneously, keeping their iterators in-sync.


tpietzsch:
thanks @etarena !

PR #157, opened 2017-03-21
Relax interface for StackView and Views.stack

### LINKED ISSUE #156: Views.stack does not work on List< Img< T > >
Currently, the following code does not compile:
```java
final List< Img< LongType > > list = new ArrayList<>();
Views.stack( list );
```
We could make it work by changing the signature of `Views.stack`
from
```java
stack( final List< RandomAccessibleInterval< T > > hyperslices )
```
 to
```java
stack( final List< ? extends RandomAccessibleInterval< T > > hyperslices )
```
The constructor of `StackView` would need to be modified accordingly.
As far as I can tell this would not be an interface-breaking change as the proposed signature is more general than the current signature. Do I miss any potential for breaking existing code here?

hanslovsky:
Using `List< ? extends RandomAccessibleInterval< T > >` instead of `List< RandomAccessibleInterval< T > >` allows for passing `List< Img< T > >` or similar.

This addresses #156 

Closes #156

### DIFF
--- a/src/main/java/net/imglib2/view/StackView.java
+++ b/src/main/java/net/imglib2/view/StackView.java
@@ -110,13 +110,13 @@ public class StackView< T > extends AbstractInterval implements RandomAccessible
 
 	private final StackAccessMode stackAccessMode;
 
-	public StackView( final List< RandomAccessibleInterval< T > > hyperslices )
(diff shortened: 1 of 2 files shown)


axtimwalde:
Thanks @hanslovskyp

tpietzsch:
👍

PR #158, opened 2017-03-31
Add Views.concatenate

hanslovsky:
`Views.concatenate` creates hyperslices along the concatenation axis and then creates a `StackView` of these hyperslices using `Views.stack`.


axtimwalde:
I am pro merging this @tpietzsch?

tpietzsch:
👍

PR #159, opened 2017-04-03
added long index based access methods with default int cast to all ac…

axtimwalde:
…cesses

This reduces code duplication for https://github.com/imglib/imglib2-unsafe/issues/1#issuecomment-291215338

### DIFF
--- a/src/main/java/net/imglib2/img/basictypeaccess/ByteAccess.java
+++ b/src/main/java/net/imglib2/img/basictypeaccess/ByteAccess.java
@@ -11,13 +11,13 @@
  * %%
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
- * 
+ *
  * 1. Redistributions of source code must retain the above copyright notice,
  *    this list of conditions and the following disclaimer.
  * 2. Redistributions in binary form must reproduce the above copyright notice,
  *    this list of conditions and the following disclaimer in the documentation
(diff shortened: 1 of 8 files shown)


hanslovsky:
I don't see any issues with this and I think this can be merged.

hanslovsky:
Actually, I thought about it a little more and I think this is not beneficial over an additional interface that extends the existing access interfaces with long index get and set methods and defaults the int index get and set methods to delegate to the long index get and set methods.

With this PR, one would need to implement 4 methods per access for long index accessing (including overwriting the default methods), whereas the interface hierarchy as it is right now in imglib2-unsafe only requires the implementaiton of 2 methods for that.

axtimwalde:
... plus the interfaces.  We can count characters :).  How about using int-based accesses in long-assuming code?  Is that something that we want?  Sorry for the premature pull request.

ctrueden:
Why not do things the other way, with the `int`-based methods being `default` and delegating to the `long` ones?

hanslovsky:
This seems like the natural choice to me as well. There is one problem with this, however: All existing accesses would need to be updated. That is why I would prefer to introduce new interfaces that extend the existing accesses and add the `long` index `get` and `set` methods and default the `int` to delegate to the `long` ones.

tpietzsch:
It doesn't look like this is supposed to be merged. Can it be closed?

PR #160, opened 2017-04-20
Add composites along arbitrary dimensions

hanslovsky:
Currently, `CompositeView` and `Views.collapse` only collapse along the last dimension ("channel axis"). It can be beneficial for caching to have the channels adjacent in memory, which would require collapsing along the first dimension. I extended the existing interface and added composition/collapsing along arbitray dimensions to `CompositeView`, `CompositeIntervalView`, and `Views.collapse`.

Once this PR is ready to be merged, I will rebase into a single commit.


hanslovsky:
I wrote a benchmark for this and, surprisingly, I did not find any benefits of a more cache-friendly accessing pattern. I will thus close this pull request.

PR #162, opened 2017-05-25
added public access to primitive types for generic integer types

axtimwalde:
deprecated redundant protected get/setValue methods

### DIFF
--- a/src/main/java/net/imglib2/type/numeric/integer/ByteType.java
+++ b/src/main/java/net/imglib2/type/numeric/integer/ByteType.java
@@ -11,13 +11,13 @@
  * %%
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
- * 
+ *
  * 1. Redistributions of source code must retain the above copyright notice,
  *    this list of conditions and the following disclaimer.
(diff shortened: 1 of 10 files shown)


tpietzsch:
> @tpietzsch if you have not objections, I would like to merge imglib/imglib2#162 ASAP
will save me four obsolete special purpose implementations for copy between primitive arrays of unsigned integers and will also make this more efficient (no double & 0xff... and cast)

Why the renaming from `setValue()` to `setByte()` etc? Why not just make `setValue()` public instead of deprecating?

axtimwalde:
@tpietzsch thanks for commenting:
1. to enable preserving the old getValue/ setValue methods in protected mode, i.e. do not change the API in terms of SemVer,
2. because it is more telling what the methods do, they are in fact all very different, no good reason to have them have the same name (at least in my eyes)

tpietzsch:
Fair enough. Then I would ask you to remember to remove the deprecated methods a few releases down the line

axtimwalde:
I promise!

axtimwalde:
I will also write some comments.  Sorry for not having done so yet.

tpietzsch:
👍  I'm fine with merging this then :-)

axtimwalde:
Great! Thanks!

PR #168, opened 2017-06-20
Clarify when coordinates are defined

### LINKED ISSUE #167: Document assumptions about when coordinates are defined
E.g.:

* For `IterableInterval`, are all coordinates within the `Interval` always iterated?
* For `RandomAccessibleInterval`, are all coordinates within the `Interval` always defined?
* How about for `RandomAccessible`? (Of course not.)
* How about `RealRandomAccessibleRealInterval`? `RealRandomAccessible`?

It would be helpful to answer this question clearly in the javadoc of each interface.

/cc @stelfrich

ctrueden:
I tried to document when each of the major accessible interfaces are defined.

However, there is one I am not sure about: `RealRandomAccessibleRealInterval`. I wrote that it is "not necessarily defined at all coordinates within the interval" but that would be in contrast to `RandomAccessibleInterval` which "is defined at all coordinates of the interval."

@tpietzsch @axtimwalde If I am wrong about this difference I'm happy to adjust the commit here; just let me know.

### DIFF
--- a/src/main/java/net/imglib2/Cursor.java
+++ b/src/main/java/net/imglib2/Cursor.java
@@ -64,6 +64,12 @@ package net.imglib2;
  * imglib Cursors, in general.
  * </p>
  *
+ * <p>
+ * It is not guaranteed that a Cursor will perform bounds checking. Asking the
+ * position or value of a Cursor that was not advanced to its first element, or
+ * was moved beyond its last element, has undefined results.
+ * </p>
+ *
  * @author Tobias Pietzsch
  * @author Stephan Preibisch
  * @author Stephan Saalfeld
--- a/src/main/java/net/imglib2/RandomAccessible.java
+++ b/src/main/java/net/imglib2/RandomAccessible.java
@@ -50,6 +50,14 @@ import net.imglib2.view.Views;
  * expect that the domain is infinite. (In contrast to this,
  * {@link RandomAccessibleInterval}s have a finite domain.)
  * </p>
+ * <p>
+ * A {@code RandomAccessible} might be defined only partially. You should be
+ * aware of this especially when creating {@link RandomAccessibleInterval}s.
+ * While it is straightforward to turn a {@code RandomAccessible} into a
+ * {@code RandomAccessibleInterval} by adding interval boundaries via
+ * {@link Views#interval}, it is your responsibility to ensure that the
+ * {@code RandomAccessible} is fully defined within these boundaries.
+ * </p>
  *
  * @author Stephan Saalfeld
  * @author Tobias Pietzsch
--- a/src/main/java/net/imglib2/RealCursor.java
+++ b/src/main/java/net/imglib2/RealCursor.java
@@ -64,6 +64,12 @@ package net.imglib2;
  * imglib Cursors, in general.
  * </p>
  * 
+ * <p>
+ * It is not guaranteed that a RealCursor will perform bounds checking. Asking
+ * the position or value of a RealCursor that was not advanced to its first
+ * element, or was moved beyond its last element, has undefined results.
+ * </p>
+ *
  * @author Tobias Pietzsch
  * @author Stephan Preibisch
  * @author Stephan Saalfeld
--- a/src/main/java/net/imglib2/RealRandomAccessible.java
+++ b/src/main/java/net/imglib2/RealRandomAccessible.java
@@ -48,6 +48,10 @@ package net.imglib2;
  * expect that the domain is infinite. (In contrast to this,
  * {@link RealRandomAccessibleRealInterval}s have a finite domain.)
  * </p>
+ * <p>
+ * Similarly to a {@link RandomAccessible}, a {@code RealRandomAccessible} might
+ * be defined only partially within the space.
+ * </p>
  * 
  * @author Stephan Saalfeld
  */
--- a/src/main/java/net/imglib2/RealRandomAccessibleRealInterval.java
+++ b/src/main/java/net/imglib2/RealRandomAccessibleRealInterval.java
@@ -43,6 +43,10 @@ package net.imglib2;
  * A function over an n-dimensional real interval that can create a random
  * access {@link Sampler}.
  * </p>
+ * <p>
+ * By convention, a RealRandomAccessibleRealInterval represents a function that
+ * is <em>defined at all coordinates of the interval</em>.
+ * </p>
  *
  * @author Stephan Saalfeld
  */
--- a/src/main/java/net/imglib2/IterableInterval.java
+++ b/src/main/java/net/imglib2/IterableInterval.java
@@ -37,6 +37,16 @@ package net.imglib2;
 /**
  * An {@link IterableRealInterval} whose elements are located at integer
  * coordinates.
+ * <p>
+ * An {@code IterableInterval} is <em>not</em> guaranteed to iterate over all
+ * coordinates of its containing {@link Interval}. In the typical case of a
+ * hyperrectangular image, it will do so; however, there are some
+ * {@code IterableInterval}s which visit only a subset of the {@code Interval}
+ * coordinates. For example, the {@code imglib2-roi} library provides means to
+ * model regions of interest (ROIs), along with the ability to iterate over
+ * coordinates within a particular ROI; see e.g.
+ * {@code net.imglib2.roi.labeling.LabelRegion}.
+ * </p>
  * 
  * @author Tobias Pietzsch
  * @author Stephan Preibisch
--- a/src/main/java/net/imglib2/RandomAccessibleInterval.java
+++ b/src/main/java/net/imglib2/RandomAccessibleInterval.java
@@ -43,6 +43,10 @@ package net.imglib2;
  * A function over an n-dimensional integer interval that can create a random
  * access {@link Sampler}.
  * </p>
+ * <p>
+ * By convention, a RandomAccessibleInterval represents a function that is
+ * <em>defined at all coordinates of the interval</em>.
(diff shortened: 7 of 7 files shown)


awalter17:
I updated the javadoc for `Cursor` and `RealCursor` to clarify bounds checking (see #170). Please feel free to let me know if you'd like any changes.

src/main/java/net/imglib2/RandomAccessible.java
```diff
@@ -50,6 +50,14 @@
  * expect that the domain is infinite. (In contrast to this,
  * {@link RandomAccessibleInterval}s have a finite domain.)
  * </p>
+ * <p>
```
axtimwalde: I do not think this is true.  Typically, a RandomAccessible is expected to be defined everywhere.  Where is this not true (while not being an errror)?

src/main/java/net/imglib2/RealRandomAccessible.java
```diff
@@ -48,6 +48,10 @@
  * expect that the domain is infinite. (In contrast to this,
  * {@link RealRandomAccessibleRealInterval}s have a finite domain.)
  * </p>
+ * <p>
```
axtimwalde: I do not think this is true.  Typically, a RealRandomAccessible is expected to be defined everywhere.  Where is this not true (while not being an errror)?

src/main/java/net/imglib2/RealRandomAccessibleRealInterval.java
```diff
@@ -43,6 +43,10 @@
  * A function over an n-dimensional real interval that can create a random
  * access {@link Sampler}.
  * </p>
+ * <p>
```
axtimwalde: I do not think this is true.  Typically, a RealRandomAccessibleInterval is expected to be defined everywhere in the interval.  Where is this not true (while not being an errror)?

src/main/java/net/imglib2/RandomAccessible.java
ctrueden: I do not know. I pulled it from the [ImgLib2 documentation on the wiki](http://imagej.net/ImgLib2_-_Accessibles#RandomAccessibles).

src/main/java/net/imglib2/RealRandomAccessible.java
ctrueden: That one I guessed, based on the statement about `RandomAccessible` not being defined everywhere.

src/main/java/net/imglib2/RealRandomAccessibleRealInterval.java
ctrueden: This one I had no idea, and actually commented in the PR here that I wasn't sure. It certainly makes sense that an RRARI actually should/must be defined everywhere within its interval. This is also more consistent with RAI.

ctrueden:
Thanks @axtimwalde for commenting.

@tpietzsch, what are your thoughts?

This discussion is making me glad we are explicitly trying to write this down; seems likely it will be beneficial in the future to avoid coding things on top of ImgLib2 with differing assumptions.

src/main/java/net/imglib2/Cursor.java
```diff
@@ -64,6 +64,12 @@
  * imglib Cursors, in general.
  * </p>
  *
+ * <p>
+ * It is not guaranteed that a Cursor will perform bounds checking. Therefore,
+ * moving a Cursor beyond the position of its last element will result in
+ * undefined behavior.
```
tpietzsch: I would modify to "...asking the position or value of a Cursor that was not advanced to its first element or was moved beyond its last element...". And possibly splitting that up into 2 sentences :-)

src/main/java/net/imglib2/RealCursor.java
```diff
@@ -64,6 +64,12 @@
  * imglib Cursors, in general.
  * </p>
  * 
+ * <p>
+ * It is not guaranteed that a RealCursor will perform bounds checking.
+ * Therefore, moving a RealCursor beyond the position of its last element will
+ * result in undefined behavior.
+ * </p>
+ *
```
tpietzsch: (same as above)
I would modify to "...asking the position or value of a Cursor that was not advanced to its first element or was moved beyond its last element...".

src/main/java/net/imglib2/RealRandomAccessibleRealInterval.java
tpietzsch: This one I would say should be expected to be defined everywhere

tpietzsch:
I think this is a difficult case. Every `RandomAccessibleInterval` is a `RandomAccessible`, so clearly there are `RandomAccessible`s that are not defined everywhere.
We might say that taking `RandomAccessible` as a method argument expresses the expectation of the argument being defined everywhere. But also this is not always true. Mostly it means just that you expect that the RA is defined where required.

With regard to the original question: `RealRandomAccessibleRealInterval` should also be expected to be defined everywhere in the interval.

ctrueden:
> I think this is a difficult case. Every `RandomAccessibleInterval` is a `RandomAccessible`, so clearly there are `RandomAccessible`s that are not defined everywhere.

Is it so difficult? A `RandomAccessible`, generally speaking, may not be defined everywhere. But a `RandomAccessibleInterval`, a specialization thereof, _is_ defined everywhere in its interval. Same for RRA and RRARI.

I have updated the two commits to reflect the comments above.

> We might say that taking RandomAccessible as a method argument expresses the expectation of the argument being defined ... where required. (For some definition of "where required" that should be made clear in the methods javadoc).

Indeed. However, this formulation seems tautological to me: a method taking a `RandomAccessible` as argument requires it to be defined in places where it requires they be defined. As usual, each method can & should provide javadoc explaining its requirements/preconditions in this regard. So I did not update the javadoc of the RA and RRA interfaces to explain this... but I can do so if you think such statements would be somehow helpful.

@stelfrich What do you think? Are the statements added by this PR now helpful enough as is? Or is further clarification needed?

And @axtimwalde, do you agree that these commits now reflect reality with respect to preconditions & expectations?

ctrueden:
> I disagree more and more with this pull request.
...
> Writing down something like 'may not be defined here and there' causes more confusion than it helps. In case that we HAVE to write something, I suggest that we guarantee nothing. But that does not seem helpful to me.

I am OK with whatever you and @tpietzsch and @StephanPreibisch feel is best with respect to expectations. However, I ___strongly___ disagree that it is unhelpful to write down the expectations. Obviously, there has been a skew with respect to those expectations. And if it were not for this PR, that skew would have remained hidden, and cause increasing problems as people continue to implement things on top of `RandomAccessibleInterval` et al with those different expectations. The javadoc _needs_ to state that either A) the values _are_ expected to be defined everywhere; or B) the caller needs to be careful because they _might not be_.

tpietzsch:
> My thinking about this is that when you use the interfaces, you make sure that you know what you are doing.

I agree with @ctrueden that we should put general expectations in the javadoc. After all, if you don't know what the interfaces are supposed to do, how should you know what you are doing.

This is helpful for users of imglib2, but it is equally important for contributors. For example putting the comment about undefined behaviour of "out-of-bounds" `Cursor` is extremely helpful if you are trying to implement an efficient `Cursor`. Between us, we may have a relatively consistent mental picture of "the rules". I think it is useful to make it explicit.

> Is it so difficult? A RandomAccessible, generally speaking, may not be defined everywhere. But a RandomAccessibleInterval, a specialization thereof, is defined everywhere in its interval. Same for RRA and RRARI respectively.

For me, this is a reasonable description of current reality. I like the descriptions like this.
@axtimwalde would it address your concerns if we add further clarification and an example to the javadoc for the non-intervals? ("In particular, `RandomAccessibleIntervals` are only guaranteed to be defined inside their interval. This is the case for example for `ArrayImg`.")

In general, there may be exceptions to things that are stated in the javadoc for specific implementations. What I don't want to happen is that we now go through all of imglib2 and slavishly make everything conform to these descriptions. But if we think about it, we try to conform to some kind of "rules" anyway. One thing that I care about very much for imglib2 is to have a small number of concepts that are consistent throughout and play nicely together. This is the reason for example, that I will put work into unifying rois (`LabelRegion`) and neighborhoods (`HypersphereShape`). These are conceptually the same thing and there is no real point in unifying them other than decreasing the number of concepts and making it more consistent. Blah, blah, blah. Long story short: I strongly endorse this PR :-)

stelfrich:
> I agree with @ctrueden that we should put general expectations in the javadoc. After all, if you don't know what the interfaces are supposed to do, how should you know what you are doing.

Since I was the trigger for @ctrueden's PR, I am in favor of putting the expectations into Javadoc. @tpietzsch is right, that deviations from those expectations should explicitly be stated in e.g. algorithms that return `RandomAccessible(Interval)`. If, however, the general expectations are not part of the Javadoc, how would an algorithm developer know when to document "deviating behavior"?

axtimwalde:
> For me, this is a reasonable description of current reality. I like the descriptions like this.
> @axtimwalde would it address your concerns if we add further clarification and an example to the javadoc for the non-intervals? ("In particular, RandomAccessibleIntervals are only guaranteed to be defined inside their interval. This is the case for example for ArrayImg.")

Fine.  But I find it missleading.  Particularly the `RA` case is unsatisfying.  E.g. `Views.extendBorder(RAI)` creates and `RA` that is defined everywhere, but this is not the case for every `RA`.  What I am trying to say tis that an `RAI` per se is not expected to be defined abnywhere.  Any specific implementation of the `RAI`, however can provide some guarantees.  This doesn't help here though.

tpietzsch:
thanks everyone!

PR #172, opened 2017-06-27
Add ceil to SubsampleIntervalView

shrucis1:
fixes subsamples not reaching last element


axtimwalde:
I am in favor of accepting this change @tpietzsch, @ctrueden, @dietzc, @StephanPreibisch, @fjug?

PR #174, opened 2017-07-17
Add RotateViewTest and fix naming

gselzer:
This pull request adds a rotation test to the Views testing suite. It checks to make sure that both the individual values and the bounds are rotated. This PR also renames permuteCoordinateInverse to permuteCoordinatesInverse to make naming consistent across methods.


ctrueden:
Thanks, @gselzer!

@tpietzsch I hope you do not mind that I merged this, since it seems like straightforward improvements to me.

PR #175, opened 2017-07-17
Fix short type casting issues

gselzer:
This pull request fixes a casting issue with `ShortType` and adds the corresponding tests.


ctrueden:
Thanks!

PR #179, opened 2017-09-20
Add TiledView and GridView

gab1one:
All work done by @MarcelWiedenmann.

### DIFF
--- a/src/main/java/net/imglib2/view/GridView.java
+++ b/src/main/java/net/imglib2/view/GridView.java
@@ -1,0 +1,347 @@
+/*
+ * #%L
+ * ImgLib2: a general-purpose, multidimensional image processing library.
+ * %%
+ * Copyright (C) 2009 - 2016 Tobias Pietzsch, Stephan Preibisch, Stephan Saalfeld,
+ * John Bogovic, Albert Cardona, Barry DeZonia, Christian Dietz, Jan Funke,
+ * Aivar Grislis, Jonathan Hale, Grant Harris, Stefan Helfrich, Mark Hiner,
+ * Martin Horn, Steffen Jaensch, Lee Kamentsky, Larry Lindsey, Melissa Linkert,
+ * Mark Longair, Brian Northan, Nick Perry, Curtis Rueden, Johannes Schindelin,
+ * Jean-Yves Tinevez and Michael Zinsmaier.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ * #L%
+ */
+
+package net.imglib2.view;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import net.imglib2.AbstractCursor;
+import net.imglib2.AbstractInterval;
+import net.imglib2.Cursor;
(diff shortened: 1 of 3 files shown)


ctrueden:
What is the plan? Are the commits going to be cleaned up? Or did you just want to let this sit?

frauzufall:
Hey! We would love to work with this, can we help somehow to get this into master and a near future maven release?

ctrueden:
@gab1one @MarcelWiedenmann Are you guys planning to get this PR to a mergeable state? Or do you need assistance?

StephanPreibisch:
Hi, this code also overlaps with implementations we did in the multiview-reconstruction, which are quite well-tested by now. I would like to discuss these in more detail before adding such methods to core.

gab1one:
Yes we originally intended to discuss this during the hackathon, I just cleaned up the commits, but @MarcelWiedenmann and @HedgehogCode and @dietzc know more about it.

@stelfrich can you point me to your methods in multiview-reconstruction? So we can check out if it covers our use cases as well?

dietzc:
@StephanPreibisch could your code be ported to imglib2 core? We need this functionality quite often, but we don't really want a dependency on multiview reconstruction (quite a big project). Maybe we can join efforts and integrate what you have in multi-view reconstruction into this PR (if something is missing) and optimize the code such that we can in the end merge it into imglib2-core?

dietzc:
@tpietzsch what about this PR? is there anything we can do to make it good enough to merge?

tpietzsch:
@dietzc I'm looking at it only now, sorry :-(
Resolving merge conflicts and fixing formatting would be a good start.

tpietzsch:
@dietzc From what I can see, `ArrangedView` is the same as `ListImg`?

gab1one:
I removed the conflict and updated the formatting.

gab1one:
> @dietzc From what I can see, ArrangedView is the same as ListImg?
In a sense yes, but `ArrangedView` con only contain RAI's and it knows about the arrangement, so that you can iterate over the arranged view as if it was one large image.

tpietzsch:
> In a sense yes, but ArrangedView con only contain RAI's and it knows about the arrangement, so that you can iterate over the arranged view as if it was one large image.

Not true.
I think you are talking about `CombinedView`?

gab1one:
Yes I was mistaken, still the arranged view knows about the grid alignment while the ListImg does not, it makes the CombinedView possible.

tpietzsch:
If by grid alignment you mean that flat list indices are arranged in a nD grid, then yes, `ListImg` also knows about grid alignment

gab1one:
Ok I took a closer look at `ListImg`, you are completely right, the Arranged View does exactly the same things, It should be easy to convert our code to use it instead and remove it.

tpietzsch:
👍

tpietzsch:
I would not put the `arrange` method in Views. I;m not sure it's needed on its own. If it is, I would put it in `ListImgs` (analogous to `ArrayImgs.bytes( final byte[] array, final long... dim )` etc. This looks like the same pattern: Here is flat data, wrap it into ListImg with the following dimensions...

Are there suggestions for better names for `CombinedView`. "Combined" is very generic, is there something that explains better what this actually does? Only thing I can think of quickly is `StitchedView`, but that also doesn't fit perfectly.

ctrueden:
Here are some words! `tile`, `mosaic`, `montage`, `grid`, `fuse`, `aggregate`, `combine`, `assemble`

I don't know the difference between "tile" vs "combine" vs "arrange" here, so I'll decline to be more specific. ;-)

frauzufall:
Or PatchedView.

gab1one:
I chose the name `GridView` and removed ArrangedView. I hope it is now in a state closer to be merged.

ctrueden:
@tpietzsch @gab1one Would it make sense to finalize this PR in person before the hackathon ends? Let me know if/how I can help.

frauzufall:
Hi all, what's the current status for this PR? Is there still work to be done or is there an alternative to this approach by now?

MarcelWiedenmann:
There is still work to be done. I believe @tpietzsch had some additional inline comments on the code, besides the ones in this thread. I can't them find anymore, though. Maybe they have been overwritten by a force push or so in the meantime.
I would gladly spend some time on this at the Dresden hackathon, but would first need to figure out what is left to do.

tpietzsch:
Here are my additional comments, dug up from a private gitter conversation and translated to english...

1.)
I would not do `Views.zeroMin()` for `TiledView`. If one wants to do that, it can be done outside. (Whereas there is no easy way to recover the min/max of a tile after it has been "`zeroMin`ed").

2.)
The added `Views` methods need javadoc.

3.)
Consider removing the `Views.grid( final List< RandomAccessibleInterval< T > > source, final long... grid )` method.
This can be simply replaced by `Views.grid( Views.arrange( ... ) )`, which seems convenient enough.

4.)
[This is a memory leak](https://github.com/imglib/imglib2/pull/179/files#diff-859cac58518c776316d0f85f6da2528fR163) if it is backed by imglib2-cache.
Those RandomAccesses potentially reference cached cells, preventing these from being garbage-collected. You should use `WeakRefs` instead. Maybe have a look at this PR, where a similar problem is fixed: https://github.com/bigdataviewer/bigdataviewer-vistools/pull/16/files

5.) TiledView should probably split in two parts:
(1) Tiling of an `Interval` into smaller intervals, and
(2) Given ein `RandomAccessibleInterval<Interval>` that you got from (1), cut these pieces from a `RandomAccessibleInterval<T>`.
Several variants of (1) could be implemented: with or without overlap, regular or not, etc, and (2) can be reused every time.

6.) _[Translator's Note: I would have to  dig into the code again to find out what I meant exactly with the following.]_ `GridView` implementation is not very efficient  but it's fine doing it like this for now.

I suggest to try to generalize and re-use elements of `CellImg`.
For example, there is [`CellGrid`](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/img/cell/CellGrid.java), which does the computations on intervals (computing min/max of a cell, translating between local, grid, and global coordinates). There is a lot of potential for re-use.

frauzufall:
Ping @MarcelWiedenmann, this would be such a useful feature :)

MarcelWiedenmann:
Oh boy - time really flies. I'll try to find some time during the coming semester break to work on it :).

StephanPreibisch:
Hi, I started looking at this ... will try to discuss it next week with @axtimwalde. We all have a lot of similar code and this is a can of worms...

ctrueden:
@tpietzsch Since you closed it, I assume something like this exists now? Could you please add a pointer to where the relevant functionality lives?

tpietzsch:
Hmm, unfortunately not. I just closed it because there has been no progress on the actual code since 2017.

tpietzsch:
Feel free to reopen. But maybe it would be better to move this to an issue?

PR #180, opened 2017-09-30
added tests for the printCoordinates method from the net.imglib2.util.Util class

testingsavvy:
The method `printCoordinates` from the class `net.imglib2.util.Util ` is not tested. It gets covered by being called by tests for other classes (e.g., `net.imglib2.img.cell.CellImgTest` and `net.imglib2.outofbounds.OutOfBoundsBorderTest` ). But the returned string is not asserted on. It is used only in the text displayed by the assertion. Therefore, any bug in the `printCoordinates` method will not be detected. We add tests to test printCoordinates exercising the branches for null and empty inputs, input with one element (not exercising the loop), and input with more than one element exercising the loop.

Similar logic applies to the rest of the functions in  the `net.imglib2.util.Util` class. We plan to continue writing tests for the rest of the class if this pull request gets in.

### DIFF
--- a/src/test/java/net/imglib2/util/UtilTest.java
+++ b/src/test/java/net/imglib2/util/UtilTest.java
@@ -1,0 +1,126 @@
+/*
+ * #%L
+ * ImgLib2: a general-purpose, multidimensional image processing library.
+ * %%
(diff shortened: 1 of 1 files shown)


ctrueden:
I think this is a great addition! :+1: @tpietzsch What do you think?

tpietzsch:
👍 thank you @testingsavvy

Issue #181, opened 2017-10-02
Fix assumption about expected length of vectors passed to setPosition and localize


axtimwalde:
While this commit

https://github.com/imglib/imglib2/commit/28c94188145cdebca845aa7569b3aa290de190f6

fixes comments in interfaces where assumptions about the expected length of vectors passed to `setPosition` and `localize` calls that was baked into assertions, it does not fix all possible implementations.  `RealPoint` and `Point` are corrected.

TODO Fix all other places.

Fix means, you cannot use `localize` to implement `setPosition` because both methods accept vectors equal or longer than the internal number of dimensions.

PR #183, opened 2017-10-03
Fix assumptions on length of passes to setPosition

gselzer:
This PR fixes all of the other instances of `setPosition` using `localizable.localize(position)` and refactors them to use
```
for ( int d = 0; d < n; d++ )
 		position[ d ] = localizable.getLongPosition( d );

```

This avoids errors when the `localizable` has dimensions that are not equal to the `position`.

See #181.


tpietzsch:
@gselzer thank you!

Issue #182, opened 2017-10-03
Enhance percentile function with quickselect


frauzufall:
I would like to make the [percentile function](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/util/Util.java#L148-L166) faster by using a [selection algorithm](https://en.wikipedia.org/wiki/Selection_algorithm) as suggested by @ctrueden. There doesn't seem to be an existing `quickselect` in Java so I could adapt [this code](http://blog.teamleadnet.com/2012/07/quick-select-algorithm-find-kth-element.html) for example. I would make a separate quickselect function and use this from inside the percentile function, or should it be the same? Should it stay here in Util or go to [imglib2-algorithm](https://github.com/imglib/imglib2-algorithm/tree/master/src/main/java/net/imglib2/algorithm)?

ctrueden:
You might want to consider `introselect` if you need to implement the algorithm yourself anyway. My vote would also be to put this into SciJava Common, since it has nothing to do with images specifically.

ctrueden:
I searched around for an introselect or quickselect implementation in Java, but it seems like the major libraries don't have one. However, with Java 8 streams + Guava, you can do something like:
```java
Streams.stream(iterable).collect(Comparators.least(k, thisComparator))
```
This leans on Guava's [Comparators.least](https://google.github.io/guava/releases/22.0/api/docs/com/google/common/collect/Comparators.html#least-int-java.util.Comparator-) function. It takes O(k) space though, as opposed to the inplace partial ordering done by the selection algorithms.

tpietzsch:
@frauzufall @ctrueden Actually, there is a quickselect in imglib2 core. Have a look at ` net.imglib2.util.KthElement`. I wrote this for building kD-Trees, IIRC. It has a ton of variants, for arrays, iterators, etc, also including option to get the permutation and apply it to another set of elements. If you need help with anything specific, I'm happy to help.

frauzufall:
Awesome! Any reason why this is not used by `net.imglib2.util.Util.percentile`? And what about Curtis' note that this is not really image specific and would therefore fit better into SciJava Common?

StephanPreibisch:
No, there is none. This is a very welcome improvement ...

frauzufall:
Alright, I am on it!

tpietzsch:
@frauzufall 

> And what about Curtis' note that this is not really image specific and would therefore fit better into SciJava Common?

[...]

PR #184, opened 2017-10-05
Percentile enhancement

frauzufall:
(-> #182)
- Modifies `percentile` function to use `KthElement.kthElement` instead of `quicksort`
- adds test for `percentile` function comparing it to `quicksort` and `Arrays.sort` (kind of redundant to the KthElement test though)
- the `quicksort (final long[] data)` function is added for consistency (just saw it while writing the test function)


tpietzsch:
👍  thank you @frauzufall!

PR #164, opened 2017-06-09
Improve ImgFactory generics

ctrueden:
See #91.

We may want to hold off on merging this until a couple of last implementation details are sorted out.

PR #185, opened 2017-10-16
Revised factory

tpietzsch:
This is a revision of how the `NativeImgFactory` mechanics works.

Every `NativeType<T>` now exposes a `PrimitiveTypeInfo` that
* provides with `getPrimitiveType` a `PrimitiveType` enum describing the backing primitive type (e.g. `BYTE`, `DOUBLE`, ...)
* provides a `createLinkedType(NativeImg<T, ? extends A>)` method to create a new instance of the `NativeType` it is attached to, linked to the given `NativeImg`.

The factories use this `PrimitiveTypeInfo` to create the correct backing `ArrayDataAccess` for images of a given type. This means that the previously public
```
public abstract NativeImg< T, ? extends ByteAccess > createByteInstance( long[] dimensions, Fraction entitiesPerPixel );
```
etc. methods of `NativeImgFactory` are gone, potentially confusing users less.
It also means that we can get rid of @axtimwalde reflection hack [`linkType(...)`](https://github.com/imglib/imglib2/blob/897b86b9ae9982dc2dc9df607bc18340ffbb3a0c/src/main/java/net/imglib2/img/cell/LazyCellImg.java#L101-L124).

Moreover, now every `NativeType` provides an enum constant describing the family of underlying access (via `getPrimitiveTypeInfo().getPrimitiveType()`).

Everything is as type-safe as before (requiring we-know-it's-correct casts in different, but not more, places). In particular, the `PrimitiveTypeInfo` that `NativeType` provides can *only* be constructed via static methods that make it impossible to create a mismatch between exposed enum constant and access family type parameter.

Main benefits:
* Provides accessible enum for access family of `NativeType`, facilitating serialisation etc.
* Will make hacks to determine said family superfluous. (imglib2-cache will simplify a lot because of this).
* Get rid of public-but-should-not-be `createByteInstance(...)` etc methods.
* Simplify factory logic, move access family type switch to only one place ([ArrayDataAccessFactory](https://github.com/imglib/imglib2/blob/98cf36ab456b1d409ff252087f1429a9f572c05b/src/main/java/net/imglib2/img/basictypeaccess/ArrayDataAccessFactory.java)). There is some fiddly method chains to make javac swallow this but no casting is necessary anywhere in the factories...

@axtimwalde @StephanPreibisch @ctrueden @dietzc Opinions?

If this is good to go, I would rebase @ctrueden PR "Improve ImgFactory generics" #164 on top of this and merge...

Bonus question: Should we get rid of `NativeImgFactory` class or keep it as a marker?


tpietzsch:
I realise that this is a big PR. To get the gist, just look at [changes to DoubleType](https://github.com/imglib/imglib2/blob/98cf36ab456b1d409ff252087f1429a9f572c05b/src/main/java/net/imglib2/type/numeric/real/DoubleType.java#L99-L105), and [ArrayImg](https://github.com/imglib/imglib2/blob/98cf36ab456b1d409ff252087f1429a9f572c05b/src/main/java/net/imglib2/img/array/ArrayImgFactory.java#L55-L69), powered by [PrimitiveTypeInfo](https://github.com/imglib/imglib2/blob/98cf36ab456b1d409ff252087f1429a9f572c05b/src/main/java/net/imglib2/type/PrimitiveTypeInfo.java) and [PrimitiveType](https://github.com/imglib/imglib2/blob/98cf36ab456b1d409ff252087f1429a9f572c05b/src/main/java/net/imglib2/type/PrimitiveType.java).

axtimwalde:
@tpietzsch I like it.  Thanks!   I am also pro removing `NativeImgFactory`.

ctrueden:
Thank you very much for working on this, @tpietzsch!

I like the direction, but was wondering about backwards compatibility. It looks like the only breaking API changes are in 98cf36ab456b1d409ff252087f1429a9f572c05b. Would it make sense to instead deprecate the old public signatures, instead of removing them? Otherwise, we'll have another bumpy migration path from imglib2 v4.x to imglib2 v5.x.

tpietzsch:
@ctrueden I'd rather remove them. This should be basically only used by someone making a new `Type` of `ImgFactory`, which is only "us" anyways.

tpietzsch:
@ctrueden I would deprecate `NativeImgFactory` because that might be actually used somewhere outside.

ctrueden:
@tpietzsch Okay. Then I think the final commit needs to bump the version to `5.0.0-SNAPSHOT` in the POM, since it breaks backwards API compatibility.

tpietzsch:
@axtimwalde @StephanPreibisch @ctrueden @dietzc @maarzt I think this is ready to go now.
I rebased and augmented @ctrueden PR "Improve ImgFactory generics" #164 on top of this in #197.
I have branches (all named 'revised-factory') following the changes and making the necessary updates in imglib2-cache, bdv-core, bdv-vistools.
While we are at it, I would like to move the Volatile types from bdv-core to imglib2, and also move them to new packages, see https://github.com/bigdataviewer/bigdataviewer-core/issues/28.
It would be great if you could have another look at this and also at #197 and let me know your opinion. I'll be busy with other things the next two weeks or so, then I'd like to merge.

PR #186, opened 2017-10-17
added a test for QuaternionToR with large numbers causing NaN and Inf/-Inf results

testingsavvy:
Added a test for the function `QuaternionToR` of the class `LinAlgHelpers`. The overflow results in NaN and Inf.


tpietzsch:
`LinAlgHelpers.quaternionToR` is for unit quaternions. So the test would test undefined behaviour.

PR #188, opened 2017-10-20
Imglib2 Loops - write loops on images: nice, simple and fast

maarzt:
An example loop: calculate grandient using central difference:
```
public static <T extends NumericType<T>> void gradientCentralDifference(RandomAccessible<T> source, RandomAccessibleInterval<T> gradient, int dimension) {

	RandomAccessibleInterval<T> backSource = Views.interval(source, Intervals.translate(gradient, -1, dimension));
	RandomAccessibleInterval<T> frontSource = Views.interval(source, Intervals.translate(gradient, 1, dimension));

	LoopBuilder.setImages(gradient, backSource, frontSource).run(
			(result, back, front) -> {
				result.set(front);
				result.sub(back);
				result.mul(0.5);
			}
	);
}
```
LoopBuilder has been carefully tuned. The resulting loop will be fast.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -225,6 +225,19 @@ Jean-Yves Tinevez and Michael Zinsmaier.</license.copyrightOwners>
 
 	<dependencies>
 		<!-- Test dependencies -->
+		<!-- JMH dependencies -->
+		<dependency>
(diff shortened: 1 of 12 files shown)


tpietzsch:
👍  thanks @maarzt

PR #189, opened 2017-10-31
add composition to Converters

axtimwalde:
I found myself repeatedly using this pattern and therefore added it to `Converters`.  The usual use-case is to compose some channels into a single channel, but in general, all per-pixel operations with n compatible type and same size inputs are possible, e.g. converting ARGB to BGRA.  Example:
```java
/* (r,g,b) to ARGBType */
List<RandomAccessibleInterval<UnsignedByteType> list = ...;
RandomAccessibleInterval<ARGBType> argb = Converters.composeReal(
  list,
  (a, b) -> {
    b.set((((
      a.get(0).get() << 16) |
      a.get(1).get() << 8) |
      a.get(2).get()) | 0xff000000);
  },
  ARGBType::new);
```

### DIFF
--- a/src/main/java/net/imglib2/converter/Converters.java
+++ b/src/main/java/net/imglib2/converter/Converters.java
@@ -35,8 +35,11 @@
 package net.imglib2.converter;
 
 import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Supplier;
 
 import net.imglib2.Cursor;
+import net.imglib2.Interval;
 import net.imglib2.IterableInterval;
 import net.imglib2.RandomAccess;
 import net.imglib2.RandomAccessible;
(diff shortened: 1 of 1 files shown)


axtimwalde:
I was also thinking about converting variants of `Views.collapse` that would work similarly.  This would be more useful where the channels are already stacked.  If we introduced that, I am not sure whether `Views` or `Converters` would be a more appropriate place.  The `Converters.compose` methods in this pull request would then directly forward to the converting `collapse` methods (as those are identical except for the missing `Views.stack`).  Not sure about the best choice.  Thoughts?

hanslovsky:
Along with this PR, I think it would be very convenient to add a converter that does exactly this (for integer types):
```java
(a, b) -> {
    b.set((((
      a.get(0).get() << 16) |
      a.get(1).get() << 8) |
      a.get(2).get()) | 0xff000000);
  }
```
For `RealType`, that converter would need to map each channel into `[0,255]` e.g.:
```java
(a, b) -> {
    b.set((((
      mapChannel1( a.get(0).get() ) << 16) |
      mapChannel2( a.get(1).get() ) << 8) |
      mapChannel3( a.get(2).get()) ) | 0xff000000);
  }
```

axtimwalde:
May be, but first things first.  It is also wrong by the way ;).  Look at the parentheses and when shifts are done (hint: my version is correct ;))

hanslovsky:
Yes, your version is correct and that is exactly why we should have the converter in there. ;)

axtimwalde:
True, this is a bug!  Overwrites alpha with 00 when writing into the blue channel... tststs...

tpietzsch:
👍  fixed it

PR #191, opened 2017-11-16
Add index-position-conversion for Interval.

hanslovsky:
Previously, it was not possible to convert indices and positions for `Interval`s with non zero min. This implements the functionality of `IntervalIndexer.indexToPositionWithOffset` and `IntervalIndexer.positionWithOffsetToIndex` but with `Interval` and `Positionable`/`Localizable` as input rather than arrays.


tpietzsch:
super useful, thanks!

PR #194, opened 2017-12-06
Add constant OutOfBounds extension for unbound generics

hanslovsky:
`OutOfBoundsConstantValue` is bound to `T extends Type< T >` as the provided extension is copied. This does not allow to extend arbitrary generics with a constant value.
I assume that the rationale for that behavior is to avoid accidental modifications of the provided type, which could actually represent meaningful data.

To get around this restriction, I add a replication of `OutOfBoundsConstantValue` and `OutOfBoundsConstantValue` that uses a `Supplier< T >` instead of a `T` to extend the `RandomAccessibleInterval` with a constant value.

### DIFF
--- a/src/main/java/net/imglib2/outofbounds/AbstractOutOfBoundsValue.java
+++ b/src/main/java/net/imglib2/outofbounds/AbstractOutOfBoundsValue.java
@@ -39,17 +39,17 @@ import net.imglib2.Interval;
 import net.imglib2.Localizable;
 import net.imglib2.RandomAccess;
 import net.imglib2.RandomAccessible;
-import net.imglib2.type.Type;
 
 /**
- * 
+ *
  * @param <T>
- * 
+ *
  * @author Stephan Preibisch
  * @author Stephan Saalfeld
  * @author Tobias Pietzsch
+ * @author Philipp Hanslovsky
  */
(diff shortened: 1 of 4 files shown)


axtimwalde:
I would prefer a solution that does not introduce a new class that does exactly the same but better than the old one.  Possibly at the cost of introducing internal API break.  But it may be possible without.

hanslovsky:
In general, I do agree. I did not want to change the existing behavior as I do not know the rationale behind it. If it is as I assumed in my initial comment I would prefer not changing it because it may result in changed behavior without failing for some code. I am thinking fo something like this:
```java
RandomAccessibleInterval< T > rai1 = ...
RandomAccessibleInterval< T > rai2 = ...
T extension = ...
extension.set( ... )
Views.extendValue( rai1, extension );
extension.set( ... )
Views.extendValue( rai2, extension );
```
This kind of code would not work as intended anymore. Instead, both `RAI`s would be extended with the same value, as no copy of the type is being made.

hanslovsky:
As of now, I do not require to `Views.extendValue` for `RandomAccessible< T >` with unbound `T`. I suggest we close this PR for now. I will not delete the branch on my fork, so we can re-visit this/re-open this PR in the future.

PR #196, opened 2018-01-09
Restore localize performance

maarzt:
The changes made in issue https://github.com/imglib/imglib2/issues/181 slow down the performance of  ArrayRandomAccess by 50%.

Aside from that  https://github.com/imglib/imglib2/issues/181 gives no reason why the changes need to be made. I think it's acceptable that `randomAccess.setPosition(localizable)` throws an ArrayIndexOutOfBounds exception, if the numbers of dimensions of `randomAccess` and `localizable` don't match?

@axtimwalde Why did you want the changes in https://github.com/imglib/imglib2/issues/181 to be made?

It's propable a good idea to revert https://github.com/imglib/imglib2/commit/28c94188145cdebca845aa7569b3aa290de190f6 too.


tpietzsch:
Interesting. I don't immediately see why this would cause such a performance hit. @maarzt Do you have an explanation?

@axtimwalde I'm for rolling it back including 28c9418. The assumption should then just be that `setPosition`, `localize`, etc always expect exactly matching `numDimensions()`.

axtimwalde:
NO!

axtimwalde:
@tpietzsch @maarzt Can we talk before you do this? I am using this assumption in my code to, without further hassle apply low dimensional transformations to higher dimensional vectors.  The assertions were testing for this condition and we simply never tested it properly.  If this comes with a performance hit, we should fix that, not this behavior.

axtimwalde:
To be more precise, I am using this to apply *in place* transformations of low dimensional transformation functions on higher dimensional vectors.

tpietzsch:
@maarzt @axtimwalde Let's find out what causes it. If we cannot fix it then I'm for fixing it in the transformation functions and rolling back. The convenience is not worth the cost in my opinion.

@axtimwalde You could fix it on the transformation side, right? Simply use the element-wise get/set calls there...

axtimwalde:
@maarzt with the new code, localize will be called less often than with the old code.  I am not sure that your benchmark https://github.com/imglib/imglib2/pull/196/commits/d82b0dee2caed1e8cb4ad4849aa1c6d577cb231d reaches the point where the method gets hot enough to be inlined and compiled.  What time are you measuring?

tpietzsch:
Here are two theories why it could be slower:

1.) Somehow polymorphism is involved and if we call `localize` we take the polymorphism hit once and for `getIntPosition` we take it once for every dimension.

2.) More likely and easy to test: The benchmark uses a `in.cursor();` which probably does some computation to get position per dimension. Simply using `in.localizingCursor();` might fix it?

axtimwalde:
I.e. I believe that this benchmark is not exclusively testing localize performance.  Maximizing the performance of this particular copy loop (which I do not think is efficient) is, in my opinion less interesting than the performance of in place transformations.  But I agree that this needs careful evaluation.  If we shift the capability of in place transformations into the transformations, we will have to change and performance test all transformations in ImgLib2 and ImgLib2 realtransform.

axtimwalde:
The Cursor theory is correct, Cursor uses the array indexer to calculate the position for each field and for tha loops over the entire array.

axtimwalde:
localizing cursor updates an internal position array and returns the location instead.

tpietzsch:
@maarzt Could you modify and run the benchmark, please?

axtimwalde:
In my opinion, this copy loop should be banned.

axtimwalde:
@maarzt, could you please also try this copy loop:

```java
final Cursor<A> a = Views.flatIterable(source);
final Cursor<A> b = Views.flatIterable(target);
while (a.hasNext())
  b.next()set(a.next());
```

axtimwalde:
Ok, I get that the other copy loop is also sometimes useful (lazy cell images), so I apologize for the request to ban it...

maarzt:
Ok, I did run some more benchmarks. Now using localizingCursor. It's hard to get any conclusions because polymorphism & just in time compilation have huge effects on the results. But these are my results:

* Without polymorphism both versions of setPosition perform equally.
* With polymorphism the older version is more efficient.

setPosition uses Localizable.localize:
```
Benchmark                                                                   Mode  Cnt          Score         Error  Units
ArrayRandomAccessPolymorphismBenchmark.copyFunctionLocalizingCursorArray   thrpt   32        377.818 ±       3.970  ops/s
ArrayRandomAccessPolymorphismBenchmark.copyFunctionLocalizingCursorCell    thrpt   32        329.030 ±      11.852  ops/s
ArrayRandomAccessPolymorphismBenchmark.copyFunctionLocalizingCursorPlanar  thrpt   32        373.717 ±       4.040  ops/s
SetPositionBenchmark.setPositionArray                                      thrpt   32  100945737.506 ± 1340462.144  ops/s
SetPositionBenchmark.setPositionPlanar                                     thrpt   32  100128303.003 ± 1240515.145  ops/s
SetPositionBenchmark.setPositionPoint                                      thrpt   32  100191000.520 ± 3678726.093  ops/s
SetPositionBenchmark.setPositionSlowCursor                                 thrpt   32   55991571.992 ±  856213.657  ops/s
```

setPostion uses Localizable.getIntPosition / getLongPosition:
```
Benchmark                                                                   Mode  Cnt         Score         Error  Units
ArrayRandomAccessPolymorphismBenchmark.copyFunctionLocalizingCursorArray   thrpt   32       277.849 ±       7.285  ops/s
ArrayRandomAccessPolymorphismBenchmark.copyFunctionLocalizingCursorCell    thrpt   32       261.162 ±       8.221  ops/s
ArrayRandomAccessPolymorphismBenchmark.copyFunctionLocalizingCursorPlanar  thrpt   32       296.853 ±      11.864  ops/s
SetPositionBenchmark.setPositionArray                                      thrpt   32  53397680.919 ± 1434640.788  ops/s
SetPositionBenchmark.setPositionPlanar                                     thrpt   32  52791813.258 ± 1762294.837  ops/s
SetPositionBenchmark.setPositionPoint                                      thrpt   32  51768996.572 ± 1396110.547  ops/s
SetPositionBenchmark.setPositionSlowCursor                                 thrpt   32  25505853.216 ±  979085.317  ops/s
```

maarzt:
In Tr2d there was some buggy code using: `randomAccess.setPosition(localizable)` where `randomAccess.numDimensions() == 3` and `localizable.numDimensions() == 2`. Which worked with an older version of imglib2 and is now broken.

ImgLib2 should only guaranty that `randomAccess.setPosition(localizable)` works, if `randomAccess.numDimensions() == localizable.numDimensions()`. Everything else makes the method setPosition harder to understand, which invites bugs, and will obfuscate the code.

@axtimwalde Please give me a pointer to the in place transformation code. I'm confident, that there is a nice solution for it.

axtimwalde:
@maarzt, I disagree.  ImgLib2 should guaranty that randomAccess.setPosition(localizable) works, if randomAccess.numDimensions() <= localizable.numDimensions().  This was the intention all the time as you can see in comments and assertions.  The implementation contained bugs that I fixed in the commit that you want to revert.  I will have to have a look at the benchmark to get a better understanding but have limited time today.  Sorry for the delay.  I will also make my case in more detail then.

axtimwalde:
BTW, who is Tr2d?

axtimwalde:
And sorry if I am confusing things unnecessarily, I did not have time to look at this in detail today or yesterday.  I just remember that I made these corrections for a good reason.

maarzt:
As far as I understand the initial goal is to "apply low dimensional transformations to higher dimensional vectors."  ? @axtimwalde Please give me an example code for this. I would like to understand it better.

Maybe there is a even nicer solution for this, which has no performance impact.

axtimwalde:
@maarzt the most recent example that triggered my cleanup is this: https://github.com/imglib/imglib2-realtransform/blob/master/src/main/java/net/imglib2/realtransform/ClippedTransitionRealTransform.java

which obviously works on `RealPositionable/Localizable`, but if it works there, it should work in discrete space too.

I am sure that there are other ways of implementing this, but my initial intention with `[Real]Localizable` and [Real]Positionable` was the current behavior.  I wrote the assertions this way and I tried hard to make it work everywhere.  Implementating `setPosition` with `localize` was a BUG and I hope I did not write it.

I am very interested in identifying the reasons for the observed performance hit and addressing and remedying it fully, but I do not agree with breaking an intended behavior just because the current implementation of that behavior has lower performance in an arbitrary performance test.

maarzt:
I always thought, randomAccessor.setPostion is implemented using Localizable.localize because that's important for the overall performance. And the copy loop I wrote made it look like there is a performance degrease of 50%. But the actual performance degrease for a more realistic copy (using a localizing cursor and a random accessor) is around 21% maybe 27%. Most likely, the performance degrease in actually useful code will be much lower. I wouldn't call this a performance hit.

But the benchmark show as well: The execution time for setPosition doubles, if Localizable.getIntPosition is used instead of Localizable.localize.

tpietzsch:
I would definitely call this a performance hit...

tpietzsch:
And it will probably get worse for higher-dimensional images. The benchmark is only 2D images.

maarzt:
@axtimwalde Please don't expect Positionable.setPosition(Localizable) to work, if the number of dimensions of the positionable and the localizable don't match. It's against (my) first intuition.

I think this makes the contraints (and the javadoc) for the setPosition method more complicated. Which is against the https://en.wikipedia.org/wiki/KISS_principle and therefor provokes future bugs.

tpietzsch:
I agree with @maarzt on this

tpietzsch:
Hmm, actually, thinking a bit more about it, I'm not so sure anymore. For both `localize(long[])` and `setPositions(long[])` it can be expected intuitively that they will just set/get the elements `0..numDimensions` from the `long[]` and therefore nothing should happen if you pass a bigger one. Behaviour with `setPosition(Localizable)` is an extension of this.

@axtimwalde There is similar expected behaviour for `localize()` I assume? It is expected to work if you pass in arguments with > numDimensions? This is missing in the javadoc.

Maybe the solution is to add a `Localizable.localizeInto(long[])` etc method, where you can pass in arguments that have less than 'Localizable.numDimensions()` and it fills them up as far as possible. Then this can be used from the `setPosition(Localizable)` call.

That being said, at the moment, I see no real downside to requiring same numDimensions everywhere and it would certainly be a clearer rule

axtimwalde:
Thanks @tpietzsch!  This was my line of thinking, it makes sense to enable the arrays passed into `setPosition` or `localize` to be longer than `numDimensions`, and so the same should be true for `RealLocalizable`.  Yes, the Javadoc needs to be corrected, I am very sorry that I forgot this when I tried to fix the same for `Positionable`.  I also agree that 20% IS a performance hit and needs to be addressed.  @maarzt I do not agree with you at this time and strongly advocate for fixing the performance hit with the current behavior.  This behavior is simple and can be well documented.

axtimwalde:
@maarzt @tpietzsch, I ran the benchmarks locally on both branches using 3D test images and the results are mixed.  Summary: `flatIterable` copy has same performance, localize copy with non-localizable `Cursor` (bad apprach) is slower in master (that is expected and fine for me), copy with `LocalizableCursor` is 30% faster in master (!).  Loop builder copy has same performance in both.  The polymorphism test for different images is mixed.  `LocalizableCursor` in master is ~20% faster for `ArrayImg` and `PlanarImg` but ~10% slower for `CellImg` (where I care most, but I did not find promising candidates at a first quick glance into `CellLocalizingCursor`).  `setPosition` tests are generally ~25% slower in master which I do not understand yet.  However, it makes me wonder if we are testing relevant performance.  Note that the `setPosition` tests perform consistently worse in master while the `LocalizableCursor` test that uses `setPosition` performs better.  It would be great if you could create a more realistic performance test such as e.g. BigDataViewer rendering (projecting) a variety of sources under transformation.

maarzt:
@axtimwalde I talked to @tpietzsch, we decided we want to do the BDV projection benchmark / or something very close to this. And when we fix the performance issue. We would like the see if there is a influence on your ClippedTransitionRealTransform.java. Can you give use an example that uses it?

maarzt:
The example BDV projection, uses `setPosition(long, int)` a lot, but never uses `setPosition(Localizable)` which leads to the following conclusions for me:
* If you guys don't know any algorithm making heavy use of `setPosition(Localizable)`, we don't need to optimize this method.
* If only `setPosition(long, int)` is used that means: It's often possible to use a higher dimensions "vector" instead of a lower dimensional one. This emphasizes @axtimwalde's arguments.

StephanPreibisch:
Hi @maarzt, I have not been following the entire time, but I definitely use setPosition(Localizable) in critical methods - except you only talk about a BDV benchmark here.

maarzt:
@StephanPreibisch I would like to benchmark an algorithm, that makes heavy use of `setPosition(Localizable)`. I would appreciate, if you can point me to such an algorithm. And maybe provide example data so that I can run it?

StephanPreibisch:
Hi, it is very simple, for example here computing a ratio between two images: https://github.com/PreibischLab/multiview-reconstruction/blob/88a344434117f380dea40e039904b8de1d603147/src/main/java/net/preibisch/mvrecon/process/deconvolution/iteration/DeconvolutionMethods.java#L46

If the two input RandomAccessibleIntervals do not have the same iteration order, one will iterate with a cursor and set a RandomAccess to its location, using randomAccess.setPositon( cursor ); 

It's a static method made for multithreading, just call it accordingly, I do it here: https://github.com/PreibischLab/multiview-reconstruction/blob/88a344434117f380dea40e039904b8de1d603147/src/main/java/net/preibisch/mvrecon/process/deconvolution/iteration/ComputeBlockThreadCPU.java#L108

Does this help?

axtimwalde:
@maarzt e.g. this code uses it https://github.com/saalfeldlab/hot-knife/blob/master/src/main/java/org/janelia/saalfeldlab/hotknife/ViewAlignedSlabSeries.java#L174-L241.  I did not have time to create a simpler example yet but it is straightforward.  You open a 3D input, make 2 different 2D affine transforms, create the transition, apply with `RealTransforms` adn show in bdv-vistools.

axtimwalde:
@StephanPreibisch do you have a chance to test the performance of your method against the imglib2 master branch or 4.6.0 and this pull request?  That would be very informative.  As you are using `localizingCursor()`, I do not expect this to be bad, but who knows...

StephanPreibisch:
Hi, I try to do this tomorrow ... does single or multithreaded matter?

maarzt:
I benchmarked the code "computing a ratio" from @StephanPreibisch, and I don't see a significant performance difference between both versions of the `setPosition(Localicable)` method. But the changes of @axtimwalde seem to make the code better suitable for JIT compiler.

I would close this PR.

maarzt:
The performance problems measured for #181 only appeared for theoretically examples. Benchmarks with real world code, didn't show any performance problems. There is no need do rollback #181.

tpietzsch:
@maarzt Thanks for investigating!!!

PR #197, opened 2018-01-15
Revised img factory

tpietzsch:
Rebase and augment @ctrueden PR #164 on top of #185


src/main/java/net/imglib2/img/ImgFactory.java
```diff
@@ -166,4 +148,84 @@
...
+	 * @return A type instance matching the type of the factory. If the given
+	 *         {@code type} argument is non-null, it is returned. Otherwise, the
+	 *         cached type instance {@link #t} is returned.
+	 */
+	@Deprecated
+	protected T cache( final T type )
+	{
+		if ( t == null ) t = type;
+		return type == null ? t : type;
+	}
```
maarzt: This type cache in ImgFactory is unnecessary complicated.
See https://github.com/maarzt/imglib2/commit/ce7966f27abab694e4b261671fe1d4f270bf19e8

src/main/java/net/imglib2/img/basictypeaccess/ArrayDataAccessFactory.java
```diff
@@ -0,0 +1,128 @@
...
+					? ( volatil
+							? ( A ) new DirtyVolatileShortArray( 0, true )
+							: ( A ) new DirtyShortArray( 0 ) )
+					: ( volatil
+							? ( A ) new VolatileShortArray( 0, true )
+							: ( A ) new ShortArray( 0 ) );
+		default:
+			throw new IllegalArgumentException();
+		}
+	}
```
maarzt: This part is not nice, and hopefully the is a way to replace it.
ArrayDataAccessFactory is a "list" of all ArrayAccess classes.
Using such list will make it harder to add ArrayAccesses in the future.
I will think about a solution tomorrow

src/main/java/net/imglib2/type/PrimitiveType.java
```diff
@@ -0,0 +1,27 @@
...
+ * <p>
+ * In conjunction with {@link AccessFlags} this describes a specific
+ * {@link ArrayDataAccess}. For example, {@code BYTE} with flags {@code DIRTY}
+ * and {@code VOLATILE} specifies {@link DirtyVolatileByteArray}.
+ * </p>
+ *
+ * @author Tobias Pietzsch
+ */
+public enum PrimitiveType
+{
```
maarzt: `PrimitiveType` is a misleading name. (Rhetorical question: Why is there no `BOOLEAN`) The name `PrimitiveType` is not telling the truth. The class is intended to represent: `ByteAccess`, `CharAccess`, `ShortAccess`, `IntAccess`, ...

Conclusion: `AccessType` of `PrimitiveAccessType` would be a better name. The constants should be named `BYTE_ACCESS`, `CHAR_ACCESS`, `SHORT_ACCESS` ...

Overall. **This class is not really needed**: `ByteAccess.class`, `CharAccess.class`, ... could be used instead. This would make the intension of the code easier to understand.

src/main/java/net/imglib2/type/PrimitiveTypeInfo.java
```diff
@@ -0,0 +1,98 @@
...
+ * @param <T>
+ *            the {@link NativeType} this is attached to
+ * @param <A>
+ *            the {@code Access} family ({@code ByteAccess},
+ *            {@code DoubleAccess}, and so on)
+ *
+ * @author Tobias Pietzsch
+ */
+public final class PrimitiveTypeInfo< T extends NativeType< T >, A >
+{
```
maarzt: 1. `PrimitiveTypeInfo` is again the wrong name. A better name would be `NativeTypeInfo` (every `NativeType` has a field of this type)
2. I would describe the purpose of this class differently:
    - The method `createLinkedType` returns a factory method to create a `NativeType`
    - The method `primitiveType` tell us which argument the factory method expects.

Conclusion: This class, represents a factory for a native type and provides some meta data, when it can be applied.
`NativeTypeFactory` would be a good name for this class.

src/main/java/net/imglib2/img/basictypeaccess/AccessFlags.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Flags that specify variants of {@link ArrayDataAccess} underlying primitive
+ * types. {@link #DIRTY} means that an access implements {@link Dirty}.
+ * {@link #VOLATILE} means that an access implements {@link VolatileAccess}.
+ *
+ * @author Tobias Pietzsch
+ */
+public enum AccessFlags
+{
+	DIRTY, VOLATILE;
```
maarzt: `AccessFlags` are only used for unused optional parameter in `ArrayDataAccessFactory`.
`imglib2-cache` is probably a better place to be for this class.

src/main/java/net/imglib2/img/basictypeaccess/AccessFlags.java
```diff
@@ -0,0 +1,56 @@
...
+		for ( final AccessFlags flag : flags )
+			if ( flag == VOLATILE )
+				return true;
+		return false;
+	}
+
+	static AccessFlags[] flags_DIRTY_VOLATILE = new AccessFlags[] { DIRTY, VOLATILE };
+	static AccessFlags[] flags_DIRTY = new AccessFlags[] { DIRTY };
+	static AccessFlags[] flags_VOLATILE = new AccessFlags[] { VOLATILE };
+	static AccessFlags[] flags_NONE = new AccessFlags[] {};
```
maarzt: Consider using EnumSet<AccessFlags> instead

tpietzsch:
I'm trying to finally finish this... I have addressed some things @maarzt mentioned above, after discussion. Here are some open questions that I would like your opinion on @axtimwalde @ctrueden @maarzt @hanslovsky:

-----

**(1.)**
@ctrueden introduced `ImgFactory` constructors with type `Supplier`s. The idea is, I think, that this should work:
```java
ArrayImgFactory<ARGBType> factory = new ArrayImgFactory<>( ARGBType::new );
```
However, as it currently is, this *doesn't* work:
```
Error:(23, 61) java: reference to ArrayImgFactory is ambiguous
  both constructor <T>ArrayImgFactory(T) in net.imglib2.img.array.ArrayImgFactory and constructor <T>ArrayImgFactory(java.util.function.Supplier<T>) in net.imglib2.img.array.ArrayImgFactory match
Error:(23, 80) java: incompatible types: cannot infer type arguments for net.imglib2.img.array.ArrayImgFactory<>
    reason: cannot infer type-variable(s) T
      (argument mismatch; net.imglib2.type.numeric.ARGBType is not a functional interface)
```
The following works
```java
ArrayImgFactory<ARGBType> factory = new ArrayImgFactory<ARGBType>( ARGBType::new );
```
but repeating the type like this is ugly, and
```java
ArrayImgFactory<ARGBType> factory = new ArrayImgFactory<>( new ARGBType() );
```
works, and is preferable.

A solution would be to remove the `ImgFactory` constructors which take a type argument.
Then the ambiguity goes away and
```java
ArrayImgFactory<ARGBType> factory = new ArrayImgFactory<>( ARGBType::new );
```
compiles.
**BUT** I have no idea how this affects use from jython, beanshell, etc scripting languages or kotlin.
Making a `Supplier` there might be difficult (if at all possible) and verbose. So I would be very reluctant to go this route.

Instead, I'm leaning towards removing the constructors with `Supplier` arguments. While it wouldn't hurt to leave them in, who would actually write
`new ArrayImgFactory<ARGBType>( ARGBType::new )`
than
`new ArrayImgFactory<>( new ARGBType() )`?

So: remove the `ImgFactory` `Supplier` constructors?

-----

**(2.)**
I'm really torn what to do about the `Volatile` accesses and types. After discussing with @maarzt, is seems good to move `AccessFlags` (`DIRTY` and `VOLATILE`) and related stuff to imglib2-cache. As mentioned https://github.com/bigdataviewer/bigdataviewer-core/issues/28 I wanted to move `VolatileARGBType` etc to imglib2-core (because it would simplify the code if they could live in the same package as `ARGBType`).

I'm leaning to move everything `Volatile` to `imglib2-cache` because it is only really useful in that context. (@axtimwalde Are there other applications of `Volatile` that I'm not aware of?)

Anyway: At the moment it is possible to create a `ArrayImg<VolatileARGBType>` but this will break at runtime. If the volatile stuff would be in `imglib2-cache` I would have less of a problem with this.
It is however also possible, to make this work, i.e,
```java
new ArrayImgFactory<>(new VolatileARGBType()).create(...)
```
can do the right thing. This would mean, however, baking `Volatile` (and `dirty`) into core for good, and making the core ImgFactories more complicated -- Although there will probably never be a use-case for `ArrayImg<VolatileARGBType>`.

I'm slightly in favor of putting everything `Volatile` into imglib2-cache to keep this mess out of imglib2-core. But this is a problem, where every solution I can think of sucks in one way or the other. Maybe you can help me pick the least sucky one... :-)

tpietzsch:
Potential work-around for (1.)
Replace `Supplier` constructors with factory methods, e.g.
```java
ArrayImgFactory<ARGBType> factory = ArrayImgFactory.for( ARGBType::new );
```

tpietzsch:
@hanslovsky then what do you think of static method `ArrayImgFactory.of( ARGBType::new )`?

tpietzsch:
@hanslovsky Ok, then I'll remove it. Less is more :-)

axtimwalde:
What's unsatisfying is that the factory actually needs a supplier, and not a type.  At the time of its emergence, there was no supplier interface though.

axtimwalde:
We have the same issue in outofbounds factories.  A clean rewrite would replace the type member by a supplier, and types would get rid of their supplier methods (createVariable, createTypeOnSam...)

hanslovsky:
> We have the same issue in outofbounds factories. A clean rewrite would replace the type member by a supplier, and types would get rid of their supplier methods (createVariable, createTypeOnSam...)

That way, `Views.extendValue` would also work for non-`net.imglib2.type.Type`.

tpietzsch:
@axtimwalde I actually thought briefly about deprecating `createTypeOnSameNativeImg()`.
Because now, there is a kind of `Supplier`, the `NativeType.getNativeTypeFactory()` (previously `PrimitiveTypeInfo`). `NativeTypeFactory.createLinkedType()` is the (supplier) factory method.

But it doesn't work out in a type-safe manner, because `NativeType` is not typed on the `ArrayDataAccess`. (And to make Volatiles work, it shouldn't be. For example, `UnsignedByteType` should be able to run on different backing `ArrayDataAccess`es: normal, dirty, volatile, dirty-volatile).

A clean rewrite is not something we will be able to realistically pull off. But it might be educational to take 2 weeks or so and get started with one. Then take lessons learned to know in which direction we should try to push existing imglib2. If we ever want to do a mini-hackathon for just this, I would be very excited... Also with Mastodon, I have now an additional perspective on how a different approach to proxy types could work. We could have a stricter separation between coordinates and types, such that a `Cursor<T>` would decompose into a `CursorCoordinate` that only cares about moving coordinates and indices, and you would explicitly get a `T` from the `Img` with a `Coordinate` (possibly reusing an existing `T` proxy). But I have no strong intuition how it would compare performance-wise. It would be exciting to play with, but a full rewrite would be so much work... especially considering the amount of stuff that has already been built on existing imglib2

tpietzsch:
@axtimwalde Regarding clean rewrite: I would consider to forego type-safety for some internals, and only presenting the a type-safe public interface. That would make many things easier -- but also potentially make imglib2 less extensible (in a type-safe manner) with new `Type`s and `ImgFactory`s etc.

tpietzsch:
Regarding **(2.)** in https://github.com/imglib/imglib2/pull/197#issuecomment-380528532
After all, I think I would just put all the `Volatile` and `Dirty` stuff into core. This way, we can at least make
```java
new ArrayImgFactory<>(new VolatileARGBType()).create(...)
```
work correctly at some point in the future (if we chose to) without breaking everything (again!) and moving classes between `imglib2-cache` and `imglib2` core.

ctrueden:
I agree with @axtimwalde about point (1): conceptually, the factory wants a `Supplier`, not a type instance. The need to preallocate the type instance is unfortunate, compared to the possibility of deferring its creation until such a type instance is actually needed.

However, regarding performance in this particular case, I'm sure there is no discernible difference anyway. And we have no acute need for the supplier, so... I'm OK with whatever is more expedient. I do really like the static creator method idea, though. 

Regarding (2), I do not feel strongly, but I do think most users of ImgLib2 never need to know about the volatile stuff at all. I liked the idea to move it to imglib2-cache since that is its primary use case. But I understand if there are technical details compelling it to remain in imglib2 core. It's all in separate packages anyway.

tpietzsch:
> But I understand if there are technical details compelling it to remain in imglib2 core. It's all in separate packages anyway.

Hmm, actually, one reason to move the Volatile Types to core was so that I could put them into the same package... But I now decided to not do that, and move everything to imglib2-cache, in separate packages.

tpietzsch:
@axtimwalde Are you ok with me also moving `Volatile` itself to imglib2-cache? And I would make it an interface while I'm at it...

axtimwalde:
I have no strong opinions, I started it outside of core and I am using imglib2-cache in every single project now, i.e. if you ask me, we could merge imglib2-core and imglib2-cache ;).

hanslovsky:
I like `Volatile` as an interface, but I do not have a strong opinion on that because
 1. It works at it is right now, and
 2. I am naively thinking that `Volatile` as an interface would reduce lines of code (e.g. `VolatileDoubleType extends DoubleType implements Volatile`) but I am not sure that that is true.

tpietzsch:
> I have no strong opinions, I started it outside of core and I am using imglib2-cache in every single project now, i.e. if you ask me, we could merge imglib2-core and imglib2-cache ;).

👍 

The reason that it's separate it that imglib2-cache has dependencies...

hanslovsky:
Just as a follow-up on what I was thinking when I said having `Volatile` as an interface might reduce code complexity:
```java
package net.imglib2.type.numeric.real;

import net.imglib2.img.NativeImg;
import net.imglib2.img.basictypeaccess.volatiles.VolatileDoubleAccess;

public class VolatileDoubleType extends DoubleType implements VolatileInterface
{

	public VolatileDoubleType( final NativeImg< ?, ? extends VolatileDoubleAccess > doubleStorage )
	{
		super( doubleStorage );
	}

	public VolatileDoubleType( final double value )
	{
		super( value );
	}

	public VolatileDoubleType( final VolatileDoubleAccess access )
	{
		super( access );
	}

	public VolatileDoubleType()
	{
		this( 0.0 );
	}

	@Override
	public boolean isValid()
	{
		if ( dataAccess == null ) 
			return false;
		return ( ( VolatileDoubleAccess ) dataAccess ).isValid();
	}

}
```
Keep in mind:
 1. This is just a naive idea and I do not know if this works or if there are certain caveats.
 2. Typing `DoubleType` on the access would avoid the dynamic cast at runtime. Obviously, this (typing on the access) is not a change that we should introduce. A workaround would be to double-track the access (this time as `VolatileDoubleAccess`) and override the relevant methods.

tpietzsch:
@hanslovsky The missing puzzle piece is the operations on the type, e.g.,
https://github.com/imglib/imglib2/blob/a87816a00e9c238ca49e2a095d01e428c6454d5e/src/main/java/net/imglib2/type/volatiles/AbstractVolatileNumericType.java#L72-L77

For example, linear interpolation of 8 `VolatileDoubleType` voxels, where one of them is invalid, should be invalid.

tpietzsch:
Argh... Ok, after trying to move it in various ways, I decided I've agonized enough over this. I will put (respectively leave) the volatile/dirty stuff in imglib2 core. I'll leave the types in a separate package so that we can move them easily later if we decide to do so.

There is already volatile stuff in core, e.g.,
`net.imglib2.converter.VolatileRealTypeARGBConverter`
`net.imglib2.display.projector.volatiles.Volatile2DRandomAccessibleProjector`
(and don't even know, if the latter class is used anywhere).
I'm putting in more classed that live in `net.imglib2....` packages in bdv-core (waiting to move for a long time), e.g.,
`net.imglib2.interpolation.randomaccess.ClampingNLinearInterpolatorRealType`
`net.imglib2.interpolation.randomaccess.ClampingNLinearInterpolatorVolatileRealType`

I would end up duplicating the core package hierarchy in imglib2-cache, e.g.,
`net.imglib2.converter` --> `net.imglib2.cache.converter`
`net.imglib2.interpolation.randomaccess` --> `net.imglib2.cache.interpolation.randomaccess`
And these classes (interpolators, converters, etc...) have nothing directly to do with caching, so I was briefly considering to make *another* artefact that has the "base cache" stuff not to do with images, so that it could be cleanly separated.

But I think this is a mess that has to wait to be fixed for another time.
Otherwise this will never get done...

PR #199, opened 2018-02-09
add generic functions as RandomAccessible and RealRandomAccessible

axtimwalde:
that generate values via side effect through a BiConsumer

I found myself implementing specialized variants of this in various projects and this to me looks like a generic convenience approach.  `PositionRandomAccessible` and `RealPositionRealRandomAccessible` in the same package are special purpose implementations of the same concept.  They could be replaced by e.g.
```java
RandomAccessible<LongType> positionRandomAccessible = new Function<LongType>(3, (x, y) -> y.set(x.getLongPosition(0)), LongType::new);
```

### DIFF
--- a/src/main/java/net/imglib2/position/AbstractFunction.java
+++ b/src/main/java/net/imglib2/position/AbstractFunction.java
@@ -1,0 +1,69 @@
+/*
+ * #%L
+ * ImgLib2: a general-purpose, multidimensional image processing library.
+ * %%
+ * Copyright (C) 2009 - 2017 Tobias Pietzsch, Stephan Preibisch, Stephan Saalfeld,
+ * John Bogovic, Albert Cardona, Barry DeZonia, Christian Dietz, Jan Funke,
+ * Aivar Grislis, Jonathan Hale, Grant Harris, Stefan Helfrich, Mark Hiner,
+ * Martin Horn, Steffen Jaensch, Lee Kamentsky, Larry Lindsey, Melissa Linkert,
+ * Mark Longair, Brian Northan, Nick Perry, Curtis Rueden, Johannes Schindelin,
+ * Jean-Yves Tinevez and Michael Zinsmaier.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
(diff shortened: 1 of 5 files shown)


axtimwalde:
I am unhappy about the names `RealFunction` and `Function`.  More distinct candidates are `BiConsumerRandomAccessible` and `BiConsumerRealRandomAccessible` or `FunctionRandomAccessible` and `FunctionRealRandomAccessible`.  Opinions @tpietzsch @ctrueden @hanslovsky @StephanPreibisch?

ctrueden:
How about `AbstractGeneratedSpace` and `GeneratedRandomAccessible` and `GeneratedRealRandomAccessible`? And make the inner `RandomAccess` implementations private since they offer no new API?

hanslovsky:
My first intuition was something along the lines of a scalar field but this is not only for scalar outputs. I like @ctrueden 's suggestion of having a `Generator` prefix. For utility methods, I suggest we would name the class `Generators`.

The term generator already has a (different) meaning in some languages, though:
https://en.wikipedia.org/wiki/Generator_(computer_programming)

axtimwalde:
Not happy.  `RandomAccessible` and `RealRandomAccessible` are already generic mappings from R^n or Z^n to T.  These two classes here are only one way of doing this in a generative way,  There are many other ways to do it in a generative way.  The particularity of this solution is that it works with BiConsumers, i.e. lambda with two input parameters via side effect.  It's not a generator because no object is generated.  Function kind of nails it but this is exactly NOT the interface that is used (that would be a generator).  I am leaning towards `BiConsumerRandomAccessible` and `BiConsumerRealRandomAccessible`.

hanslovsky:
We should add a JavaDoc remark to the constructor that the `typeSupplier.get()` parameter should create distinct objects if multiple accesses are generated down the line, e.g. in multi-threaded environments.

tpietzsch:
@axtimwalde I like `FunctionRandomAccessible` better than `BiConsumerRandomAccessible`. It describes the intended purpose better. That the function must be specified as a `BiConsumer` because of imglib peculiarities is not so important in my opinion.

It might be good to add a constructor with `Supplier<BiConsumer<...>>`. This would allow for `BiConsumer`s with internal state (temporary variables ...). Every access could have its own `BiConsumer` instance then. This could also be added later, but I would put it in the javadoc then, that the `BiConsumer` must be thread-safe.

axtimwalde:
Excellent suggestions @tpietzsch, much appreciated, as always.  I will do it so and then merge immediately.

tpietzsch:
@axtimwalde Are you still working on this?

axtimwalde:
Ahem... I should... Will do in the coming days and then merge.  Thanks for the reminder...

PR #200, opened 2018-03-03
 Integer ShearTransforms with configurable shearing factor

keithschulze:
Currently, ShearTransform only allows n-dimensional image to be sheared along the shear dimension, with respect to the reference dimension, by a factor of 1. This PR is a refactor of the ShearTransform to allow transform with a `shearFactor` other than 1.

```java
// forward transform
c[ shearDimension ] = c[ shearDimension ] + c[ referenceDimension ] * shearFactor

// backward/inverse transform
c[ shearDimension ] = c[ shearDimension ] - c[ referenceDimension ] * shearFactor 
// or c[ shearDimension ] = c[ shearDimension ] + c[ referenceDimension ] * -shearFactor

```

Tries to avoid any breaking changes to the existing API, consisting of classes for forward ShearTransform, backward InverseShearTransform and AbstactShearTransform with common functionality; however, introducing a shearFactor simplifies implementation, since the inverse transformation simply requires the inverse of the `shearFactor`. Hence, basically all functionality is now in the AbstractShearTransform class, while ShearTransform and InverseShearTransform just provide basic constructors that ensure the `shearFactor` is set correctly. Perhaps it might be worth deprecating the InverseShearTransform and AbstractShearTransform classes (and then removing them in the next major version), and just moving all functionality to ShearTransform?

Views have also been updated to include static methods that allow a shearFactor to be specified.

### DIFF
--- a/src/main/java/net/imglib2/transform/integer/shear/AbstractShearTransform.java
+++ b/src/main/java/net/imglib2/transform/integer/shear/AbstractShearTransform.java
@@ -11,13 +11,13 @@
  * %%
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
- *
+ * 
  * 1. Redistributions of source code must retain the above copyright notice,
  *    this list of conditions and the following disclaimer.
  * 2. Redistributions in binary form must reproduce the above copyright notice,
  *    this list of conditions and the following disclaimer in the documentation
  *    and/or other materials provided with the distribution.
- *
+ * 
  * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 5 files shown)


axtimwalde:
@keithschulze, did you make a performance evaluation?  Replacing an addition by addition + multiplication is a good candidate for slowing things down in the trivial case.

keithschulze:
@axtimwalde I did some benchmarking with JMH when I implemented this for [llsm](https://github.com/monash-merc/llsm), but that was quite some time ago (>1 year) and I can't remember if I compared it to the exisiting code. I'll dig out those benchmarks and do the comparison. I imagine you are right that there will be a slow down.

hanslovsky:
I am in favor of deprecating `InverseShearTransform` and `AbstractShearTransform`.
Also very interested in the benchmark results.

PR #201, opened 2018-03-12
Constant accesses

hanslovsky:
Constant access always return the same value.


tpietzsch:
What is the purpose?

hanslovsky:
For propagation of painted data to higher resolution scales in multi-scale painting, I would like to be able to set complete blocks to a constant value without touching every single pixel. Ideally, this can also be persisted to disk with disk cached cell imgs.

hanslovsky:
I am currently trying to cache these constant cells to disk, and I noticed that the hierarchy should be slightly different.

tpietzsch:
I think it would be better to put these classes into your own project until it has proven useful? Then migrate to `imglib2` once it settled. The same way as the `Volatile` accesses started in bdv...

hanslovsky:
I can live with that.

tpietzsch:
👍

PR #202, opened 2018-04-12
Note that randomAccess() to access outside Interval is a bug

### LINKED ISSUE #171: Consider introducing Views.isDefined()
Introducing `Views.isDefined(RandomAccessibleInterval, Interval)` would enable Ops to check if a `RandomAccessibleInterval` needs an extension for a computation to succeed.

Currently, filter-like Ops (that might access pixels outside of the `Interval`) can be provided an optional `OutOfBoundsStrategy` for convenience. If none is provided, we apply a default one without checking if that's actually necessary. This overwrites an already existing `OutOfBoundsStrategy`. While you can always set the `oobf` to `null` when executing an Op on an already extended image, we want to be more intelligent about when to extend an image.

Work happening on https://github.com/imglib/imglib2/tree/defined-bounds

**See also**
- https://github.com/stelfrich/imglib2/commits/random-access-defined-bounds-no-interfaces
- [AbstractNeighborhoodBasedFilter.java#L53-L55](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/filter/AbstractNeighborhoodBasedFilter.java#L53-L55)

/cc @tpietzsch

ctrueden:
It's helpful to be explicit about how `randomAccess(Interval)` (and `realRandomAccess(RealInterval)`) must be used to access out-of-bounds values of RAI and RRARI.

### DIFF
--- a/src/main/java/net/imglib2/RandomAccessibleInterval.java
+++ b/src/main/java/net/imglib2/RandomAccessibleInterval.java
@@ -47,6 +47,15 @@ package net.imglib2;
  * By convention, a RandomAccessibleInterval represents a function that is
  * <em>defined at all coordinates of the interval</em>.
  * </p>
+ * <p>
+ * There is no guarantee regarding whether the function is defined outside the
+ * bounds of its interval. If the function is known to be defined for
+ * out-of-bounds values in a particular interval, the
+ * {@link #randomAccess(Interval)} method should be used to access those
+ * values&mdash;whereas the {@link #randomAccess()} (no arguments) method <em>is
+ * not intended to access out-of-bounds values</em>. See
+ * {@link RandomAccessible#randomAccess()} for related discussion.
+ * </p>
  *
  * @author Stephan Saalfeld
  */
--- a/src/main/java/net/imglib2/RealRandomAccessibleRealInterval.java
+++ b/src/main/java/net/imglib2/RealRandomAccessibleRealInterval.java
@@ -47,6 +47,15 @@ package net.imglib2;
  * By convention, a RealRandomAccessibleRealInterval represents a function that
  * is <em>defined at all coordinates of the interval</em>.
  * </p>
+ * <p>
+ * There is no guarantee regarding whether the function is defined outside the
+ * bounds of its interval. If the function is known to be defined for
+ * out-of-bounds values in a particular interval, the
+ * {@link #realRandomAccess(RealInterval)} method should be used to access those
+ * values&mdash;whereas the {@link #realRandomAccess()} (no arguments) method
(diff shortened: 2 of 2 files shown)


tpietzsch:
Isn't this already exhaustively discussed in https://github.com/imglib/imglib2/blob/a87816a00e9c238ca49e2a095d01e428c6454d5e/src/main/java/net/imglib2/RandomAccessible.java#L75-L106?

If we want to feature this more prominently on the `RandomAccessibleInterval` javadoc, I would just copy this over...

(And, I still think that it might be good to change the default behaviour. But of course that doesn't mean that the current behaviour should not be thoroughly documented).

ctrueden:
@tpietzsch OK, shall I just put a link from the RAI to the RA javadoc then?

tpietzsch:
Or just copy & paste that section? Both are fine

ctrueden:
Added a link.

In practice, is this usage of `randomAccess(Interval)` where the `Interval` argument is a padded version of the bounded interval, a way forward you would recommend here? In cases where the bounded interval is _not_ a rebounded version of something extended, then `ArrayIndexOutOfBoundsException` will be thrown when attempting to access OOB samples. We are left with the conundrum of how to tell that an input RAI is valid in the desired padded range...

tpietzsch:
I still don't like how it is formulated.

The comment https://github.com/imglib/imglib2/blob/a87816a00e9c238ca49e2a095d01e428c6454d5e/src/main/java/net/imglib2/RandomAccessible.java#L75-L106 *explicitly* lays out a case where *it is ok to access the RAI outside the interval*.

It should more clearly state that you should not use a RA obtained by `randomAccess()` outside the interval of the RAI, and you should not use a RA obtained by `randomAccess(interval)` outside `interval`.

tpietzsch:
> We are left with the conundrum of how to tell that an input RAI is valid in the desired padded range...
This is what #171 is about.

I think the discussion of all these related subjects is something that we could more efficiently have in person in the upcoming hackathon? @hanslovsky @axtimwalde you and me will all be there...

hanslovsky:
I won't be but I will see you afterwards when you are at Janelia. I think I will be represented by @axtimwalde in that matter as our views on this are very similar if not the same.

tpietzsch:
Ok, sorry @hanslovsky, somehow I missed that you are not there

ctrueden:
> I still don't like how it is formulated. The comment ... explicitly lays out a case where it is ok to access the RAI outside the interval.

Good point, @tpietzsch. I made another attempt at formulating it, now that I have a better understanding of the preconditions of the different methods.

tpietzsch:
I like it!

hanslovsky:
I like it as well.

PR #203, opened 2018-04-16
RandomAccessibleIntervalCursor: test if we need to switch line before calling randomAccess.fwd()

gab1one:
In most cases randomAccess.fwd(0) should be very cheap, but if we
operate on a CellImg, this can cause a cell to be loaded unnecessarily.


dietzc:
image

for the sake of completeness. Benchmark performed with `RandomAccessibleIntervalCursorBenchmark`. 20 iterations and 40 runs. Master vs. changed branch.

PR #207, opened 2018-05-08
Add Views.moveAxis and Converters.mergeColor

maarzt:
This PR add:

* Converters.mergeColor creates a color view of and image of UnsignedByteType. The view supports read&write.
* Views.moveAxis: Very useful to create a color view, if the axis order needs to be changed before Converters.mergeColor can be applied.
* Utility class Localizables: Methods that are very useful to write test, or during prototyping.


ctrueden:
With the addition of `Converters.mergeARGB`, should we deprecate the `CompositeXYProjector`?

PR #208, opened 2018-05-12
Add the infrastructure for boolean[]-backed images

ctrueden:
Previously, one could extend the type system to have a `NativeBoolType` (plus `BoolAccess` and `BoolArray`) externally, and use it with `ArrayImg` and/or `PlanarImg`. However, as of ImgLib2 v5, the `NativeTypeFactory` was locked down tightly, as a final class with only a private constructor. Hence, creating a `NativeTypeFactory` that produced `NativeBoolType` instances became impossible.

This branch addresses the issue by introducing the needed classes to work with `boolean` arrays in the same way as other primitive arrays.

For an example of where this is used, see https://github.com/ctrueden/game-of-life.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
 	</parent>
 
 	<artifactId>imglib2</artifactId>
-	<version>5.5.1-SNAPSHOT</version>
+	<version>5.6.0-SNAPSHOT</version>
 
 	<name>ImgLib2 Core Library</name>
 	<description>A multidimensional, type-agnostic image processing library.</description>
--- a/src/main/java/net/imglib2/img/array/ArrayImgs.java
+++ b/src/main/java/net/imglib2/img/array/ArrayImgs.java
@@ -34,12 +34,14 @@
 
 package net.imglib2.img.array;
 
+import net.imglib2.img.basictypeaccess.BooleanAccess;
 import net.imglib2.img.basictypeaccess.ByteAccess;
 import net.imglib2.img.basictypeaccess.DoubleAccess;
 import net.imglib2.img.basictypeaccess.FloatAccess;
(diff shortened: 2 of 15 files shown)


tpietzsch:
@ctrueden I think this would be a good addition to make it complete (no matter how useful it is in practice). It makes me slightly uncomfortable that my introduction of `NativeTypeFactory` takes away from extensibility. So, I would definitely add the `PrimitiveType.Boolean`, `BooleanAccess`, etc infrastructure.

Whether the actual `NativeBoolType`, and `ArrayImgs.booleans(...)` etc should be added is more debatable. I tend to adding them. Do you have an opinion, @axtimwalde ?

I tried to rebase it onto current master an ran into the issue that now (with #205) I need to define byte count for `PrimitiveType.BOOLEAN`. Unfortunately, there is no `java.lang.Boolean.BYTES`. In theory the number of bytes is implementation (JVM) dependent, although in practice, I think it is `1` everywhere. So should we just put `1`?

axtimwalde:
I agree with everything @tpietzsch said.  Thanks @ctrueden!

hanslovsky:
`BitType` and `BoolType` could be modified to use `ByteAccess` internally. No public signatures would require changes as far as I can tell, as long as we introduce a `LongAccessAsByteAccess` for use in `BitType`. That would also unify `BitType`, `BoolType`, and `NativeBoolType`.

I saw you introduced volatile `ByteAccess`es. Currently, we cannot create any volatile `BooleanType` (#209) but it is probably useful to have them in case the interface changes in the future.

ctrueden:
I have rebased over the latest master to resolve the conflicts. I used 1 for the boolean byte count.

PR #211, opened 2018-06-06
Add methods to test equality of (Real)Localizables

ctrueden:
The equals method of individual Localizable and RealLocalizable implementations may impose different or additional requirements beyond merely equality of position. These utility methods offer a consistent way to verify that two such objects have the same position. They will be used in imglib2-roi; see imglib/imglib2-roi#35.


ctrueden:
Rebased over the latest master.

PR #214, opened 2018-06-21
Add copy helper methods for accesses

hanslovsky:
This is useful for me in the imglyb context, copying from contiguous `numpy.ndarray` to imglib2 `ArrayDataAccess`.

### DIFF
--- a/src/main/java/net/imglib2/img/basictypeaccess/Accesses.java
+++ b/src/main/java/net/imglib2/img/basictypeaccess/Accesses.java
@@ -1,0 +1,136 @@
+/*
+ * #%L
+ * ImgLib2: a general-purpose, multidimensional image processing library.
+ * %%
+ * Copyright (C) 2009 - 2018 Tobias Pietzsch, Stephan Preibisch, Stephan Saalfeld,
+ * John Bogovic, Albert Cardona, Barry DeZonia, Christian Dietz, Jan Funke,
+ * Aivar Grislis, Jonathan Hale, Grant Harris, Stefan Helfrich, Mark Hiner,
+ * Martin Horn, Steffen Jaensch, Lee Kamentsky, Larry Lindsey, Melissa Linkert,
+ * Mark Longair, Brian Northan, Nick Perry, Curtis Rueden, Johannes Schindelin,
+ * Jean-Yves Tinevez and Michael Zinsmaier.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ * #L%
+ */
+
+package net.imglib2.img.basictypeaccess;
+
+/**
+ * Utility and helper methods for accesses ({@link ByteAccess} etc).
+ *
+ * @author Philipp Hanslovsky
+ */
+public interface Accesses
+{
+
+	public static void copy( final ByteAccess source, final ByteAccess target, final int size )
+	{
+		copy( source, target, 0, size );
+	}
+
+	public static void copy( final ByteAccess source, final ByteAccess target, final int start, final int stop )
+	{
+		for ( int index = start; index < stop; ++index )
+		{
+			target.setValue( index, source.getValue( index ) );
+		}
+	}
+
+	public static void copy( final CharAccess source, final CharAccess target, final int size )
+	{
+		copy( source, target, 0, size );
+	}
+
+	public static void copy( final CharAccess source, final CharAccess target, final int start, final int stop )
+	{
+		for ( int index = start; index < stop; ++index )
+		{
+			target.setValue( index, source.getValue( index ) );
+		}
+	}
+
+	public static void copy( final DoubleAccess source, final DoubleAccess target, final int size )
+	{
+		copy( source, target, 0, size );
+	}
+
+	public static void copy( final DoubleAccess source, final DoubleAccess target, final int start, final int stop )
+	{
+		for ( int index = start; index < stop; ++index )
+		{
+			target.setValue( index, source.getValue( index ) );
+		}
+	}
+
+	public static void copy( final FloatAccess source, final FloatAccess target, final int size )
+	{
+		copy( source, target, 0, size );
+	}
+
+	public static void copy( final FloatAccess source, final FloatAccess target, final int start, final int stop )
+	{
+		for ( int index = start; index < stop; ++index )
+		{
+			target.setValue( index, source.getValue( index ) );
+		}
+	}
+
+	public static void copy( final IntAccess source, final IntAccess target, final int size )
+	{
+		copy( source, target, 0, size );
+	}
+
+	public static void copy( final IntAccess source, final IntAccess target, final int start, final int stop )
+	{
+		for ( int index = start; index < stop; ++index )
+		{
+			target.setValue( index, source.getValue( index ) );
+		}
+	}
+
+	public static void copy( final LongAccess source, final LongAccess target, final int size )
+	{
+		copy( source, target, 0, size );
+	}
+
+	public static void copy( final LongAccess source, final LongAccess target, final int start, final int stop )
+	{
+		for ( int index = start; index < stop; ++index )
+		{
+			target.setValue( index, source.getValue( index ) );
+		}
+	}
+
+	public static void copy( final ShortAccess source, final ShortAccess target, final int size )
+	{
+		copy( source, target, 0, size );
+	}
+
+	public static void copy( final ShortAccess source, final ShortAccess target, final int start, final int stop )
+	{
+		for ( int index = start; index < stop; ++index )
+		{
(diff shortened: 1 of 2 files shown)


src/main/java/net/imglib2/img/basictypeaccess/Accesses.java
```diff
@@ -0,0 +1,136 @@
...
+ */
+public interface Accesses
+{
+
+	public static void copy( final ByteAccess source, final ByteAccess target, final int size )
+	{
+		copy( source, target, 0, size );
+	}
+
+	public static void copy( final ByteAccess source, final ByteAccess target, final int start, final int stop )
```
ctrueden: This signature assumes the offset into `source` and `target` will be the same, which may not necessarily hold. I suggest modeling the API after `System.arraycopy`, which supports separate offsets for the source and dest arrays. So in this case: `copy(ByteAccess src, int srcPos, ByteAccess dest, int destPos, int length)`.

src/main/java/net/imglib2/img/basictypeaccess/Accesses.java
```diff
@@ -0,0 +1,136 @@
...
+	{
+		for ( int index = start; index < stop; ++index )
+		{
+			target.setValue( index, source.getValue( index ) );
```
ctrueden: In cases where we have an `ArrayDataAccess`, we can do better performance-wise, by calling `getCurrentStorageArray()` followed by `System.arraycopy`. I suggest adding this as a special case somehow.

src/main/java/net/imglib2/img/basictypeaccess/Accesses.java
hanslovsky: Good point. I wonder how complext this should be (you could even think of a different stride). Maybe it's best to just have `copy( source, target, size )`, then.

src/main/java/net/imglib2/img/basictypeaccess/Accesses.java
hanslovsky: I wonder if this is pre-mature optimization. If copying from one access to another access is a bottleneck, there is probably something wrong with the code. Also, `ArrayDataAccess.getCurrentStorageArray()` returns `Object` and is not even guaranteed to return array.
 A way around this would be to add
```java
public default void copyInto( ByteAccess target )
{
// do the copy as in the proposed methods
}
```
to the `ByteAccess` etc. interfaces and override with special case implementations for `ByteArray` etc.

hanslovsky: FWIW, I made a little "benchmark":
```java
import net.imglib2.img.basictypeaccess.ByteAccess;
import net.imglib2.img.basictypeaccess.array.ByteArray;

public class ByteAccessCopyTest
{

	public static void main( final String args[] )
	{
		final int iterations = 100000;

		final int size = 1000000;

		final ByteArray source = new ByteArray( size );
		final ByteArray target = new ByteArray( size );

		for ( int i = 0; i < iterations; ++i )
		{
			copy( source, target, size );
		}

		final long tStartAccessCopy = System.nanoTime();
		for ( int i = 0; i < iterations; ++i )
		{
			copy( source, target, size );
		}
		final long tStopAccessCopy = System.nanoTime();
		final long tAccessCopy = tStopAccessCopy - tStartAccessCopy;

		for ( int i = 0; i < iterations; ++i )
		{
			System.arraycopy( source.getCurrentStorageArray(), 0, target.getCurrentStorageArray(), 0, iterations );
		}

		final long tStartArraycopy = System.nanoTime();
		for ( int i = 0; i < iterations; ++i )
		{
			System.arraycopy( source.getCurrentStorageArray(), 0, target.getCurrentStorageArray(), 0, iterations );
		}
		final long tStopArraycopy = System.nanoTime();
		final long tArraycopy = tStopArraycopy - tStartArraycopy;

		System.out.println( "access copy: " + tAccessCopy * 1.0 / iterations );
		System.out.println( "array  copy: " + tArraycopy * 1.0 / iterations );

	}

	private static final void copy( final ByteAccess source, final ByteAccess target, final int size )
	{
		for ( int i = 0; i < size; ++i )
		{
			target.setValue( i, source.getValue( i ) );
		}
	}

}
```

```
access copy: 111722.59448
array  copy: 4472.66502
```
Might be worth having those special case implementations. Is there a good way to check if an `Object` is instance of primtiive type other than doing `obj instance of byte[] || obj instance of char[]` etc.?

ctrueden: I expanded your performance test to include `ShortArray` and `FloatArray` as well mixed into one scope, and speed remains equivalent to `System.arraycopy`. Pretty impressive JIT!

src/main/java/net/imglib2/img/basictypeaccess/Accesses.java
ctrueden: My preference would be to match the `System.arraycopy` method. It will be familiar to Java programmers, and is more powerful.

tpietzsch: would be my preference too

hanslovsky: Will mimic the `System.arraycopy` method then.

hanslovsky:
I updated the interface to follow `System.arraycopy`. I c&p'ed and modified the relevant parts of the `System.arraycopy` documentation. Is that an issue?

hanslovsky:
Would it be useful to add a method
```java
copy( Object src, int, Object dest, int, int )
```
that picks the right `copy` method or throws if `src` and `dest` are not the same kind of access?

src/main/java/net/imglib2/img/basictypeaccess/Accesses.java
```diff
@@ -0,0 +1,451 @@
...
+			final int srcPos,
+			final ByteAccess dest,
+			final int destPos,
+			final int length )
+	{
+
+		if ( src == dest )
+		{
+			if ( srcPos == destPos ) { return; }
+			if ( srcPos < destPos )
```
tpietzsch: This looks incomplete. What if `destPos > srcPos`

src/main/java/net/imglib2/img/basictypeaccess/Accesses.java
```diff
@@ -0,0 +1,451 @@
...
+			{
+				for ( int index = length - 1; index >= 0; --index )
+				{
+					dest.setValue( index + destPos, src.getValue( index + srcPos ) );
+				}
+			}
+			return;
```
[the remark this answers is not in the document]
hanslovsky: Good catch, the `return` should be within the inner `if` block. This should resolve this comment and the other one, as well. Will add appropriate test cases to make sure all three cases are handled.

src/main/java/net/imglib2/img/basictypeaccess/Accesses.java
hanslovsky: Addressed in e4897a5

maarzt:
The PR introduces a lot of code duplication. Which could by completely removed.
One way would to add an `CopyElement` interface to all the accesses.
```
interface CopyElement<A> {
   public copyElement(int srcIndex, A destAccess, int destIndex);
}
```
Accesses.copy would only be needed once for this CopyElement class

hanslovsky:
Then the code duplication would reside within each of the basic access interfaces (`ByteAccess`, `CharAccess`, etc.) instead of a single class because there are no c++ like templates in Java, and on top of that we would have introduced another interface. I don't see any benefits in that.

maarzt:
Yes, Java generics are barely usable at all that's true :rage:. But that's no excuse :angel:. Despite all the pain, super cool stuff like for example imglib2 can be done :heart:. That reminds me, `ByteType` enables fast access to the elements of `ByteAccess`. And similar stuff is possible for `IntAcess`, `ShortAccess`... We could use that, to kill all the code duplication:
https://gist.github.com/maarzt/77ee5ad4de5a380ad95daf4e3bbae9ca

maarzt:
It would be totally useful to have an interface, maybe called `BasicAccess`, that's implemented by all accesses. I find it very wired, that `ByteAccess` and `IntAccess` share no common interface.
The `BasicAccess` interface could have such insanely useful methods like `public int size()`.

And one could write generic methods like:
```
public <A extends BasicAccess> void copy(A source, int sourceIndex, A destination, int destIndex, int size);
```

hanslovsky:
The code in your gist looks interesting and would certainly reduce code duplication. What if `src` and `dest` are different implementations of the same interface, e.g. `SomeByteAccess src` and `SomeOtherByteAccess dest` -- would that still compile?

As for a common interface for all accesses, that is probably not within the scope of this PR.

hanslovsky:
I implemented a more generic interface to reduce code duplication as suggested by @maarzt 
This comes at the cost of:
 - support for `CharAccess`
 - compile time safety as now any `Object` can be passed for `src` and `dest` (exception is thrown if incompatible combination of `src`, `dest` is passed)

hanslovsky:
What are the opinions on this?

> This comes at the cost of:
> - support for CharAccess
> - compile time safety as now any Object can be passed for src and dest (exception is thrown if incompatible combination of src, dest is passed)

StephanPreibisch:
Hi, while this is useful I was wondering if this should be in core ... maybe we should have something like imglib2-utils? What do you think?

Some things from imglib2/src/main/java/net/imglib2/util/ could also go there ... including quite some copy methods in there.

Cheers,
Stephan

maarzt:
I would suggest to put it into `imglib2-algorithms`

hanslovsky:
I disagree. I think that imglib2 core is a good place for something as simple as a copy and as closely tied to the access interfaces. Of course, we could modularize more, but I think that, if we go that route, we will find many things that could go into their own separate modules, e.g.
 - imglib2-convert
 - imglib2-views
 - imglib2-transform
 - imglib2-awt

That's mostly a matter of taste.

imagejan:
> That's mostly a matter of taste.

It's also a matter of dependency hierarchy, no? Would any consumers of `imglib2` benefit of a more modular component setup? Are there (many) projects that would require `imglib2` but _not_ `imglib2-algorithm` (or any of the potential sub-projects, for that matter) and therefore benefit from a smaller footprint?

maarzt:
To me it feels like: This will lay around, convolute the code, and not being used much. An imglib2 user is not intended to do sophisticated actions on Accesses. Imglyb is a very specific (useful) usecase.
@hanslovsky What am I missing? Why do you think others would need it as part of imglib2 core?

hanslovsky:
> An imglib2 user is not intended to do sophisticated actions on Accesses. 

That is just plain wrong. There is different levels of expertise, and some will deal with accesses directly.

There is no algorithmic element to this, so no reason to put this into imglib2-algorithm.

maarzt:
StephanPreibisch wrote:
> ... maybe we should have something like imglib2-utils? What do you think?

If imglib2 has such a diverse set of users (beginners that script in FIJI to n5 and imglyb). It's maybe realy beneficial if there is imglib2-expert-utils, such that a less experienced user isn't confused by all the wild range of functionality. Or imglib2-backend-utils?

axtimwalde:
@maazt @hanslovsky I agree with @hanslovsky that this is a core/ utils component because it does not introduce dependencies, has no side effects, and enables core to be used in a way that is necessary.  On the other hand, I do not see any use for it outside of imglyb currently, and it does not require changes in core for it to work, so I would keep this code in imglyb until the need arises to use it elsewhere.  This will also give it time to generalize/ modularize better and possibly support other types.  We have been following this pattern elsewhere, e.g. BDV.  @maarzt I do not share the sentiment that this code would "lay around and convolute the code" because it does not intersect with other aspects of core.

axtimwalde:
@imagejan I am using imglib2 without algorithm all the time but would not benefit from excluding stuff from the existing imglib2 core which is rather small.  Our previous approach about modularization was along dependency lines.  This is the only reason why imglib2-realtransform is not in imglib2 core.

hanslovsky:
The consesus seems to be that this adds not enough usabiltiy to imglib2 core, gonna close.

hanslovsky:
For now, people can use imglib2-imglyb to use `Accesses.copy`:
https://github.com/imglib/imglib2-imglyb/blob/c6dda7d07808e860e2f727cb2708fa1a5878462f/src/main/java/tmp/net/imglib2/img/basictypeaccess/Accesses.java
Right now I do not see a good way to create a static method that is specialized for each access type (array access, unsafe access, etc.) without adding interface (methods). We should probably re-visit if there is ever a need outside imglyb.

PR #217, opened 2018-06-27
BitType: Fix #216 by setting entitiesPerPixel to 1/64

### LINKED ISSUE #216: Default construction of an ArrayImg<BitType> creates overly large storage arrays
See https://github.com/gab1one/imglib2-bittype for a demonstration of this issues, which lets an `ArrayImg<BitType>` to take up as much memory as an `ArrayImg<LongType>`.

gab1one:


src/main/java/net/imglib2/type/logic/BitType.java
```diff
@@ -326,7 +326,7 @@ public String toString()
 	@Override
 	public Fraction getEntitiesPerPixel()
 	{
-		return new Fraction();
+		return new Fraction(1, 64);
```
ctrueden: Awesome catch!

Could you please add a regression test that calls `ArrayImgs.bits` and then rips out the `long[]` (via `getCurrentStorageArray()`) and asserts that it is the right length?

gab1one: https://github.com/gab1one/imglib2-bittype/blob/master/src/test/java/org/knime/knip/ArrayImgOversizeTest.java Like this one?

ctrueden: Something like that, yeah! Makes sense to add it here, right?

src/test/java/net/imglib2/type/logic/BitTypeArrayImgSizeTest.java
```diff
@@ -0,0 +1,81 @@
...
+public class BitTypeArrayImgSizeTest
+{
+
+	@Test
+	public void testOverSizedBitImageExplicitStorageArray()
+	{
+
+		final int numLongs = ( int ) Math.ceil( 1000 * 1000 / 64 );
+
+		final ArrayImg< BitType, LongArray > img = ArrayImgs.bits( new LongArray( new long[ numLongs ] ), 1000, 1000 );
```
imagejan: Would it make sense to also test with image sizes that are not divisible by 64? That is, `1000 * 1000 / 64 == 15625`, so we don't actually test "partially filled" `long`s. How about a 100 * 100 image (`1000 * 1000 / 64 == 156.25`)?

gab1one: Good point @imagejan I modified the test :+1:

tpietzsch:
Good catch @gab1one! Please format with imglib2 code style. Then it's good to go from my point of view.

PR #219, opened 2018-07-19
Asserts And Radomized Images for JUnit Tests with ImgLib2

maarzt:
When writing unit tests that involved ImgLib2, I often find it to hard:
* To check if two images are equal.
* To generate a randomized image as input for my unit test.

This PR adds this needed functionality.

### DIFF
--- a/src/main/java/net/imglib2/test/ImgLib2Assert.java
+++ b/src/main/java/net/imglib2/test/ImgLib2Assert.java
@@ -1,0 +1,140 @@
+/*
+ * #%L
+ * ImgLib2: a general-purpose, multidimensional image processing library.
+ * %%
+ * Copyright (C) 2009 - 2018 Tobias Pietzsch, Stephan Preibisch, Stephan Saalfeld,
+ * John Bogovic, Albert Cardona, Barry DeZonia, Christian Dietz, Jan Funke,
+ * Aivar Grislis, Jonathan Hale, Grant Harris, Stefan Helfrich, Mark Hiner,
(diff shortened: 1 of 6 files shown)


ctrueden:
See also https://github.com/imagej/imagej-ops/blob/imagej-ops-0.41.1/src/test/java/net/imagej/ops/AbstractOpTest.java

maarzt:
Hallo Florian,

Ich konnte da Problem reproduzieren. Es gibt einen einfachen Workaround:
Im "labkit segmentation" Tab, musst du ERST unter Segmentation den "Classifier-#2"
auswählen, dann die zwei Striche malen und dann auf train clicken.

Wenn der "Classifier-#2" nicht ausgewählt wurde, entsteht der Fehler den du
beschrieben hast. (Das ist ein Bug)

Tracking funktioniert mit Max component size = 4000
und max(0, (size-area))^2 = 0

Viele Grüße,
Matthias

----- Original Message -----

maarzt:
> See also https://github.com/imagej/imagej-ops/blob/imagej-ops-0.41.1/src/test/java/net/imagej/ops/AbstractOpTest.java

I had a closer look into that class, not everything there can be part of this commit. Opening images for example as imglib2-core supports no file io. But I added:
* Support for UnsignedVariableBitLengthType
* A method to check if two images are equal
* A method asDoubleArray to convert an image of RealType to double[]

I'm unsure how the assertIterableEquals could become part of this PR?

tpietzsch:
Cool, very useful!

PR #220, opened 2018-07-19
LoopMath: an approach to avoid explicit loops for pixel-wise operations

acardona:
LoopMath: an approach to avoid explicit loops for pixel-wise operations
across multiple RandomAccessInterval instances.

Currently supports:
Add, Sub, Mul, Div, Min, Max, Neg, Abs.

This approach is particularly useful for scripting languages with terrible
performance when looping, such as jython.

An example piece of code is in the test.

### DIFF
--- a/src/main/java/net/imglib2/loops/LoopMath.java
+++ b/src/main/java/net/imglib2/loops/LoopMath.java
@@ -1,0 +1,438 @@
+package net.imglib2.loops;
+
+import java.util.Iterator;
+import java.util.LinkedList;
+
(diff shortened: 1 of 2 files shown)


acardona:
An improvement would be to implement the Max, Min as new net.imglib2.type.operators, along with Add, Sub, Mul, Div, and then be able to perform all math with the NumericType operators. If this is welcome--I really think is the proper way forward--, I'm happy to do it for all types that implement NumericType (a lot!).

PR #221, opened 2018-07-19
LoopMath: an approach to avoid explicit loops for pixel-wise operations across multiple RandomAccessInterval instances.

acardona:
LoopMath: an approach to avoid explicit loops for pixel-wise operations
across multiple RandomAccessInterval instances.

Currently supports:
Add, Sub, Mul, Div, Min, Max, Neg, Abs.

These functions are implemented using the Type's add, sub, mul, div and compareTo methods,
from the NumericType interfaces from net.imglib2.type.operators.

This approach is particularly useful for scripting languages with terrible
performance when looping, such as jython.

An example piece of code is in the test.

### DIFF
--- a/src/main/java/net/imglib2/loops/LoopMath.java
+++ b/src/main/java/net/imglib2/loops/LoopMath.java
@@ -1,0 +1,438 @@
+package net.imglib2.loops;
+
+import java.util.Iterator;
+import java.util.LinkedList;
+
+import net.imglib2.IterableRealInterval;
+import net.imglib2.RandomAccessibleInterval;
+import net.imglib2.type.numeric.RealType;
+import net.imglib2.view.Views;
+
+/**
+ * An easy yet relatively high performance way to perform pixel-wise math
+ * on one or more {@link RandomAccessibleInterval} instances.
+ * 
+ * <pre>
+ * {@code
+ * RandomAccessibleInterval<A> img1 = ...
+ * RandomAccessibleInterval<B> img2 = ...
+ * RandomAccessibleInterval<C> img3 = ...
+ * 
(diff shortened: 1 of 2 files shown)


kephale:
Just pointing out that the new Op classes in src/main/java/net/imglib2/loops/LoopMath.java might be easily confused with imagej-ops, e.g. https://github.com/imagej/imagej-ops/tree/master/src/main/java/net/imagej/ops/math

axtimwalde:
Cute and useful.  I understand the need for something simple to use, but not ready for pull-request and may be not in this repository, therefore closing.  Overlaps significantly with ImageJ-Ops https://nbviewer.jupyter.org/github/imagej/tutorials/blob/master/notebooks/ImageJ_Tutorials_and_Demo.ipynb but without any harder stuff considered, e.g. compatibility test rejects non-iterable RAIs but does not consider iteration order, i.e. does not check correct property.  Wouldn't it be possible to build this on top of loop-builder which makes efforts to trick the JVM into efficient looping in case of polymorphisms?  Looping several times with unary and binary operators and generating intermediate results shouldn't necessarily be slower than looping over everything simultaneously because of (possible) data locality?

acardona:
It does consider iteration order: see line 90, and lines 109-130.

Cannot be built on top of the LoopBuilder: the function, if defined in e.g. jython, will not execute efficiently.

The LoopMath summarizes everything that was good about the imglib-scripting, and respects the Type-based math completely.

If anything, all that I find missing is the ability to provide a Converter from I (inputs) to O (outputs), which I will implement for completion, with a default one using the current approach (setReal(getRealDouble()).

PR #222, opened 2018-07-24
Let LoopBuilder Support Up To Six Images

maarzt:
I think it's time to support more than three images in LoopBuilder
as the topic came up in the ImageJ forum:
http://forum.imagej.net/t/imglib2-any-way-to-apply-a-math-operation-to-three-randomaccessibleinterval-and-store-the-result-in-a-fourth-without-an-explicit-loop/12079


tpietzsch:
👍

PR #224, opened 2018-07-31
Fix Views.concatenate for axis other than last dimension

### LINKED ISSUE #223: ConcatenateView not functioning as expected
As shown in [this notebook](https://github.com/imagej/tutorials/blob/op-directory/notebooks/1_-_Using_ImageJ/Ops/transform/concatenateView.ipynb), `concatenate` does not output an image as expected. Firstly, it would be expected that a call to `concatenate` with `concatenationAxis = 0` would produce a landscape image, with all of the images placed side by side. However the image returned by `concatenate` within the notebook. is oriented vertically, which makes no sense. Note that the output also appears to be mirrored, as there is no way that the ramp image can be rotated to align with the `concatenate` output.

Going into `ConcatenateViewTest` shows these errors: if we pretend we are combining the first two images within the notebook (with the first image being 200 px by 150 px and the second image being 400 px by 150 px) by changing [`dims`]{https://github.com/imglib/imglib2/blob/82f3676367b2c0fe73d5948b27bdc3e90ccaf705/src/test/java/net/imglib2/view/ConcatenateViewTest.java#L58) to `{600, 150}` and `divider` to `200`, we get an `AssertionError`. If we comment out the lines providing the assertion error, and below these lines add:

```
System.out.println( Arrays.toString( Intervals.dimensionsAsIntArray( img ) ) );
System.out.println( Arrays.toString( Intervals.dimensionsAsIntArray( concatenated ) ) );
```

we see that `concatenated` and `img` do not have the same dimensions

hanslovsky:
Quoting my comment from #223

`Views.concatenate` uses `Views.stack` internally, which stacks images only along the last dimension. This could probably be solved by

 - first permuting the last dimension and the concatenation dimension, then
 - stacking all hyperslices along the last dimensions, and finally
 - un-doing the permutation of last dimension and concatenation dimension

Also added a test that concatenates along the first dimension (failed before fix, runs successfully now)

Fixes #223


hanslovsky:
Follow-up on argument validation:
In general, it seems that imglib2 does very little to no argument checks and pre-conditions are checked via `assert` statements in debug mode (or rather `-ea` flags). For this specific case, we could think of three illegal argument cases (besides `null` passed as any of the arguments, which could be checked via `Objects.requireNonNull`):
 - `sources.size() < 1` as indicated by [`assert sources.size() > 0`](https://github.com/imglib/imglib2/blob/b1e6831c1228c713287a240056470ff204c52184/src/main/java/net/imglib2/view/Views.java#L1516)
 - incosistent dimensions of the RAIs in `sources`
 - `concatenationAxis < 0` or `concatenationAxis 

If these checks are to be added, they should not throw a checked `Exception` as not to change the interface (which would require another major version bump for imglib2).

Maybe @tpietzsch @axtimwalde @StephanPreibisch can comment

axtimwalde:
I'd stick with testing via assertions to limit the runtime cost in real applications.  We've done this elsewhere in ImgLib2 to minimize redundant runtime testing.  An, in my eyes positive, side-effect of this is that ImgLib2 is incredibly hackable, i.e. you can do stuff that wasn't necessarily intended without paying a penalty.  E.g. stacking RAIs of variable sizes and then intervaling the result should work :).

hanslovsky:
Thanks for clarification @axtimwalde 
I will not add parameter checks, then and this is good to merge, from my side.

src/test/java/net/imglib2/view/ConcatenateViewTest.java
```diff
@@ -57,19 +64,60 @@
 
 	private final long[] dim = new long[] { 3, 4, 5, 6 };
 
+	private final long[] revertedDim = new long[] { 6, 5, 4, 3 };
```
maarzt: There is no reason to have `dim`, `revertedDim`, `divider`, `axis` and `revertedAxis` as fields. They should be declared as variables in the methods testConcatenate and testConcatenateFirstAxis.

The prefix `reverted` is kind of confusing. Because it rises the idea that some reverting happens. But that's actually not the case. Ok the order of the numbers are reverted, but that has no meaning for the test.

hanslovsky: Addresed in f7d1841, passing values as anonymous variables now.

hanslovsky:
Thanks for benchmarking @maarzt 
I wasn't sure about the performance impact and unaware of `Views.moveAxis`. I agree that we should go for simpler implementation if there is no performance penalty. Will squash.

tpietzsch:
@maarzt Do you still have the benchmark around?

maarzt:
No I didn't keep it. It was kind of boring. I compared the two working versions of Views.concatenate.:
* Version by @hanslovsky which uses Views.permute twice
* And the version that uses Views.moveAxis
Both had exactly the same performace.

tpietzsch:
Ok. I just was thinking about putting a bit more work into this. Cutting the input up into hyperslices and then stacking them is not optimal... (This had nothing to do with this fix, which is fine, and I completely believe they have the same performance 😄)

Issue #130, opened 2016-04-25
Should Type implement a valueEquals method?


axtimwalde:
I keep finding myself in places where I want to test whether two samples have the same value.  In the conventional Java world, the `Object.equals(Object)` method would be the right thing to use, unfortunately, this is not well suited for ImgLib2 because (a) we are explicitly not interested in equivalency on the reference level, (b) `Object.equals(Object)` is not typed and would thus require dynamic casting, (c) the JDK uses `Object.equals(Object)` in other contexts (e.g. `HashSet`) that would trigger potentially unintended side-effects.  Overriding `Object.equals(Object)` requires that we think about how to override `Object.hashCode()` which is not currently my intention.

I would introduce a single method interface `type.operators.ValueEquals` and see how far I can got from there.  If you have comments or red flags that I cannot see right now, please let me know.

tpietzsch:
@axtimwalde As we would say in german: "Das Kind ist schon in den Brunnen gefallen"
https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/type/numeric/real/AbstractRealType.java#L132-L148
I don't remember who introduced it originally and why. I just remembered @ctrueden PR to make it more consistent. If @axtimwalde would rather remove it, maybe you two can discuss options...

axtimwalde:
@tpietzsch, oh well...  It doesn't hurt to have that in parallel.  We do, however, lack a method to generically match the values of e.g. `ARGBType` and `RealType`.  I will give it a shot.

tpietzsch:
@axtimwalde sounds good. BTW the "axtimwalde was assigned by tpietzsch just now" was some keyboard shortcut I pressed mistake and I don't know how to revert it...

ctrueden:
@tpietzsch If you push the `?` key it will tell you.

axtimwalde:
@tpiezsch thanks for the trust assignment, makes me feel trusted ;).

axtimwalde:
I opened a pull request https://github.com/imglib/imglib2/pull/132

PR #225, opened 2018-08-09
Add ARGBType::equals

maarzt:
There's some surprising behavior with imglib2 types:
* `new IntType(5).equals(new IntType(5)) == true`
* `new ARGBType(5).equals(new ARGBType(5)) == false`

This is because ARGBType misses a implementation vor Object::equals.

This PR does the following:
* Add proper implementations for Object::equals to all NumericTypes, including ARGBType.
* Add unit test, to test all implementations of NumericType for their basic functionality.

### DIFF
--- a/src/main/java/net/imglib2/type/numeric/complex/AbstractComplexType.java
+++ b/src/main/java/net/imglib2/type/numeric/complex/AbstractComplexType.java
@@ -35,6 +35,8 @@
 package net.imglib2.type.numeric.complex;
 
 import net.imglib2.type.numeric.ComplexType;
+import net.imglib2.type.numeric.real.DoubleType;
+import net.imglib2.util.Util;
 
 /**
  * TODO
@@ -169,23 +171,28 @@ public abstract class AbstractComplexType< T extends AbstractComplexType< T >> i
 	}
 
 	@Override
-	public boolean equals( final Object o )
+	public boolean valueEquals( T other )
 	{
-		if ( !getClass().isInstance(o) )
+		return DoubleType.equals( getRealDouble(), other.getRealDouble() ) &&
+				DoubleType.equals( getImaginaryDouble(), other.getImaginaryDouble() );
+	}
+
+	@Override
+	public boolean equals( final Object obj )
+	{
+		if ( !getClass().isInstance(obj) )
 			return false;
 		@SuppressWarnings("unchecked")
-		final T t = (T) o;
-		return getRealDouble() == t.getRealDouble() &&
-			getImaginaryDouble() == t .getImaginaryDouble();
+		final T t = (T) obj;
+		return AbstractComplexType.this.valueEquals( t );
 	}
 
 	@Override
 	public int hashCode()
 	{
-		// NB: Compute similar hash code to java.lang.Double#hashCode().
-		final long rBits = Double.doubleToLongBits(getRealDouble());
-		final long iBits = Double.doubleToLongBits(getImaginaryDouble());
-		return (int) (rBits ^ (rBits >>> 32) ^ iBits ^ (iBits >>> 32));
+		final int rHash = Double.hashCode( getRealDouble() );
+		final int iHash = Double.hashCode( getImaginaryDouble() );
+		return Util.combineHash( rHash, iHash );
 	}
 
 	@Override
--- a/src/main/java/net/imglib2/type/label/BasePairBitType.java
+++ b/src/main/java/net/imglib2/type/label/BasePairBitType.java
@@ -52,7 +52,49 @@ public class BasePairBitType extends AbstractBit64Type< BasePairBitType > implem
 	// hom many bits a BasePairBitType contains
 	private static final int NBITS = 3;
 
-	public static enum Base { gap, N, A, T, G, C, U; }
+	public enum Base
+	{
+		gap( ' ' ),
+		N( 'N' ),
+		A( 'A' ),
+		T( 'T' ),
+		G( 'G' ),
+		C( 'C' ),
+		U( 'U' );
+
+		private final char c;
+
+		Base( final char c )
+		{
+			this.c = c;
+		}
+
+		public char getChar()
+		{
+			return c;
+		}
+
+		public static Base fromChar( final char c )
+		{
+			switch ( c )
+			{
+			case 'A':
+				return A;
+			case 'T':
+				return T;
+			case 'G':
+				return G;
+			case 'C':
+				return C;
+			case 'U':
+				return U;
+			case ' ':
+				return gap;
+			default:
+				return N;
+			}
+		}
+	}
 
 	// this is the constructor if you want it to read from an array
 	public BasePairBitType( final NativeImg< ?, ? extends LongAccess > bitStorage )
@@ -119,25 +161,7 @@ public class BasePairBitType extends AbstractBit64Type< BasePairBitType > implem
 	@Override
 	public byte baseToValue()
 	{
-		final Base base = get();
-
-		switch ( base )
-		{
-		case N:
-			return 1;
-		case A:
-			return 2;
-		case T:
-			return 3;
-		case G:
-			return 4;
-		case C:
-			return 5;
-		case U:
-			return 6;
-		default:
-			return 0;
-		}
+		return ( byte ) get().ordinal();
 	}
 
 	@Override
--- a/src/main/java/net/imglib2/type/label/BasePairBitType.java
+++ b/src/main/java/net/imglib2/type/label/BasePairBitType.java
@@ -92,29 +92,13 @@ public class BasePairBitType extends AbstractBit64Type< BasePairBitType > implem
 	@Override
 	public Base get()
 	{
-		return Base.values()[ (int)getBits() ];
+		return Base.values()[ ( int ) getBits() ];
 	}
 
 	@Override
 	public int compareTo( final BasePairBitType c )
 	{
-		final Base input = get();
-		final Base compare = c.get();
-
-		if ( input == compare )
-		{
-			return 0;
-		}
-		switch ( input )
-		{
-			case gap: return -1;
-			case N: return compare == Base.gap ? 1 : -1;
-			case A: return compare == Base.gap || compare == Base.N ? 1 : -1;
-			case T: return compare == Base.G || compare == Base.C || compare == Base.U ? -1 : 1;
-			case G: return compare == Base.C || compare == Base.U ? -1 : 1;
-			case C: return compare == Base.U ? -1 : 1;
-			default: return 1;
-		}
+		return get().compareTo( c.get() );
 	}
 
 	@Override
--- a/src/main/java/net/imglib2/type/label/BasePairBitType.java
+++ b/src/main/java/net/imglib2/type/label/BasePairBitType.java
@@ -49,18 +49,21 @@ import net.imglib2.type.NativeTypeFactory;
  */
 public class BasePairBitType extends AbstractBit64Type< BasePairBitType > implements BasePairType< BasePairBitType >
 {
+	// hom many bits a BasePairBitType contains
+	private static final int NBITS = 3;
+
 	public static enum Base { gap, N, A, T, G, C, U; }
 
 	// this is the constructor if you want it to read from an array
 	public BasePairBitType( final NativeImg< ?, ? extends LongAccess > bitStorage )
 	{
-		super( bitStorage, 3 );
+		super( bitStorage, NBITS );
 	}
 
 	// this is the constructor if you want it to be a variable
 	public BasePairBitType( final Base value )
 	{
-		super( value.ordinal() );
+		super( value.ordinal(), NBITS );
 	}
 
 	// this is the constructor if you want it to be a variable
--- a/src/main/java/net/imglib2/type/label/BasePairCharType.java
+++ b/src/main/java/net/imglib2/type/label/BasePairCharType.java
@@ -124,54 +124,13 @@ public class BasePairCharType extends AbstractNativeType< BasePairCharType > imp
 	@Override
 	public void set( final Base base )
 	{
-		switch ( base )
-		{
-		case A:
-			setChar( 'A' );
-			return;
-		case T:
-			setChar( 'T' );
-			return;
-		case G:
-			setChar( 'G' );
-			return;
-		case C:
-			setChar( 'C' );
-			return;
-		case U:
-			setChar( 'U' );
-			return;
-		case gap:
-			setChar( ' ' );
-			return;
-		default:
-			setChar( 'N' );
-			return;
-		}
+		setChar( base.getChar() );
 	}
 
 	@Override
 	public Base get()
 	{
-		final char value = getChar();
-
-		switch ( value )
-		{
-		case 'A':
-			return Base.A;
-		case 'T':
-			return Base.T;
-		case 'G':
-			return Base.G;
-		case 'C':
-			return Base.C;
-		case 'U':
-			return Base.U;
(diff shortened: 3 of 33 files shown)


tpietzsch:
we consciously did not add `equals()` implementations to all types, although I don't remember the details. There is `valueEquals()`.
See discussions in https://github.com/imglib/imglib2/pull/132 and https://github.com/imglib/imglib2/issues/130
Maybe @axtimwalde has more pointers

maarzt:
I look into #132 and #130, and summerize:
1. It's nice to have `ValueEquals` because it is typed and does not need casts like `equals`, and it makes no implications on `hashCode`.
2. Implementing `equals` makes it necessary to implement `hashCode`, otherwise the are surprises with java collections. 
3. The current situation is inconsistent because `AbstractRealType` implements `equals` but other types don't. As @tpietzsch said "Das Kind ist in den Brunnen gefallen." (pour child)

Based on that:
* I don't like the inconsistency (see 3. and the PR description), which already causes bad surprises to me.
* To make it consistent, I would (as already done) add implementation for `equals` and (TODO) `hashCode`.

maarzt:
Here is the commit that implemented equals for AbstractRealType:
https://github.com/imglib/imglib2/commit/054e9a9479706995e88816f0e0720d549d97cf59

axtimwalde:
Types typically being proxies to some underlying access stretches the hashCode equality and consistency requirements of `Object.equals` (https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals(java.lang.Object).  It also requires casting, so I introduced `valueEquals` which is a better match with what I would use it for.  Not sure how necessary an `Object.equals` implementation would be.  The way you implemented it, it mimics `valueEquals` with a dynamic cast.  Is this useful?

tpietzsch:
I would rather not introduce it, to encourage use of `valueEquals`.
Maybe to make it consistent we should make equals mean Object identity for all Types?

ctrueden:
> Maybe to make it consistent we should make equals mean Object identity for all Types?

That would be my vote.

maarzt:
I think the implementation of Type::equals makes better simpler code possible. (We should implement it better simpler code is the overall purpose of imglib2) A use case illustrates this point came to my mind:

Lets assume we have two ImageJ ops of the same class. And we want to see if the parameter lists are equal. This is needed for a "new fancy caching ops results" implementation. To see if the two parameter lists are equal, we element wise apply `Object::equals`. This approach greatly works is we have implemented equals in the imglib2 types but it becomes way more difficult if we don't.

maarzt:
> @axtimwalde wrote:
>Types typically being proxies to some underlying access stretches the hashCode equality and consistency requirements of Object.equals (https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals(java.lang.Object).

Yes, the proxy object thingy is a bit confusing. The proxies destination changes and equals suddenly returns a different value. **But** that's true for all modifiable Java objects, for example Lists, Sets .... People are very much used to it. And that's why this is **explicitly allowed** in the javadoc above:
> It is consistent: for any non-null reference values x and y, multiple invocations of x.equals(y) consistently return true or consistently return false, **provided no information used in equals comparisons on the objects is modified.**

Reassigning the proxies destination is a modification to information used in the equals comparison, and that's fine. The equals result might change but this is not considered inconsistent.

tpietzsch:
> To see if the two parameter lists are equal, we element wise apply Object::equals

I think this is illustrates my earlier point.
If you get two lists of `Type<T>` then you first should check whether they have the same type type `T` and then use `valueEquals()` to not have an unnecessary per element type-check.

Having `equals` encourages the more expensive approach.

maarzt:
I could find some code that would break, if the implementation of RealType::equals changes to Object::equals:
https://github.com/maarzt/imglib2-labkit/blob/master/src/main/java/net/imglib2/labkit/labeling/Labelings.java#L85

maarzt:
I think we should add the implementation for `equals` and `hashCode`. Otherwise this would be a missing functionality. @tpietzsch also agreed with the following argument:

Imglib2 does a great effort and introduces `IntergerType`, `DoubleType` and all the others. Because Java's primitive types and boxed types just don't work for image processing algorithms. It's imglib2's goal to enable fast and generic algorithms in Java. And it does a great job. **But what if such generic algorithm needs a `Set` or a `Map`???**... It would be super nasty without implementations of `hashCode` and `equals`.

(The user is forced to replace the missing `hashCode` and `equals` method. Write wrappers for all types. Or write a `Map` that could be used for all imglib2 types. Both tasks are to difficult for many users.
Most users will just implement, stop writing generic algorithm or do a nasty hack, that breaks way to often, and has bad performance. )

maarzt:
@axtimwalde What do you think, regarding the last post?

axtimwalde:
Sure, go ahead :).  We should still encourage folks to use `valueEquals` wherever possible.  BTW, as you are raising the convenience of using standard data structures.  We could never implement collections with images because everything in the collections framework depends on entries being object instances.  Now that there are streams, the contracts are different and the API is more interesting and flexible.  Have you considered implementing meaningful stream support for images, for parallel mapping and accumulation and stuff?

hanslovsky:
> But what if such generic algorithm needs a Set or a Map???

Ideally, we should have proxy object based sets and maps in imglib (does Mastodon have those?). This would probably be the cleanest solution for storing imglib2 `Type` and could be a nice student project.
Otherwise, we may face
 - bad performance because of the [`Type.copy`](https://github.com/maarzt/imglib2-labkit/blob/master/src/main/java/net/imglib2/labkit/labeling/Labelings.java#L90) that is necessary for consistency (if we have to store many objects in a collection), and/or
 - confusion about values inside a map that magically change because users may not be aware of the implications of proxy objects and do not store copies of the proxy object.

I personally use efficient [GNU Trove](https://bitbucket.org/trove4j/trove) collections for sets/maps of number but I am also aware that, for convenience, some prefer to use standard Java collections.

That being said, I agree that `hashCode` and `equals` should have consistent behavior, favoring the behavior that introduces the least amount of broken code.

maarzt:
> Ideally, we should have proxy object based sets and maps in imglib2

I totally agree.

maarzt:
I think this PR has been thoroughly discussed. And is ready for merge.

tpietzsch:
> Ideally, we should have proxy object based sets and maps in imglib (does Mastodon have those?)

Yes,  has them (also based on [GNU Trove](https://bitbucket.org/trove4j/trove)).
They are called [`RefCollection`](https://github.com/bigdataviewer/mastodon-collection/blob/07fd6c3340221a4557cde0bffc79a25e348e119c/src/main/java/org/mastodon/collection/RefCollection.java)s. There is `RefList`, `RefObjectMap`, `RefRefMap`, etc.
There are also `RefCollection` wrappers for standard `java.util` collections, so everything can be written generically for both proxy and "normal" objects.

However, I think it would not be straightforward to adapt to imglib2.

The `RefCollection`s rely on not mixing proxy objects from different "pools" in the same collection (because objects are stored in the collection by their pool index, and the collection has one "source" pool.). That would analogously mean that you can only have `Type`s from the same `Img` in a proxy based collection, which would be not enforceable in practice. So there would need to be an encoding that includes the `Img` in the primitive. We would need to use `long`s for proxy indices, because it is easy to have >2G pixels (pixels are small, maybe just 1 bit...). In contrast, objects are large (8-byte aligned) and therefore an `ArrayList` typically stores compressed object references that are just `int`s. So, the actual proxy list might consume more memory than an `ArrayList` with compressed references (not counting the objects of course). I agree that the `Type.copy` is typically much worse, e.g., `UnsignedByteType.copy()` creates at least 56 bytes worth of objects, to represent 1 byte... But if you put the same copy into enough `ArrayList`s it might still be cheaper than `long` proxy indices. So as always, there are trade-offs...

I think for storing `Type`s in collections we would rather want something like value types.
A 1D `Img<UnsignedByteType>` is basically a `List<UnsignedByteType>`... Except it has a fixed size etc... But this would be my preferred solution. Value type collections. Interface-wise it would come very close to the mastodon `RefCollections`, but semantics would be slightly different.

tpietzsch:
> I think this PR has been thoroughly discussed. And is ready for merge.

Not thoroughly enough, not yet. I still have something to nitpick :-)

For the volatile types, `equals()` violates the contract https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#equals-java.lang.Object- because it is not an equivalence relation (it is not reflexive)
https://github.com/imglib/imglib2/blob/5ef5fa9e52997f34d3bf24d14d7f8f69ac228709/src/main/java/net/imglib2/type/volatiles/AbstractVolatileNumericType.java#L127-L137
`equals()` falls back on `valueEquals()` which is not reflexive (not required by the `ValueEquals` contract currently, but I think it should be).

I  would fix this by changing `valueEquals()` to
```
	public boolean valueEquals( T other )
	{
//		return ( isValid() && other.isValid() ) && t.valueEquals( other.t );
		return ( isValid() == other.isValid() ) && t.valueEquals( other.t );
	}
```
Any objections @axtimwalde?

Also, just so it doesn't get lost: Yesterday I merged @ctrueden `NativeBoolType`, this needs to override `equals` and `hashCode` too. (I can do that on master after merging this PR).

axtimwalde:
@tpietzsch not sure about the valueEquals change.  My thinking about this was that `VolatileType`s that are invalid behave like NaN, i.e. NaN != NaN which nobody seems to consider a violation of equivalence.  But that could also be tested by checking `isValid()`, so practically, doing it your way `(isValid == isValid) && (value == value)` is probably the more useful approach.  I can live with both.

tpietzsch:
Wow, crazy: https://docs.oracle.com/javase/8/docs/api/java/lang/Double.html#equals-java.lang.Object-
For `double` NaN != NaN
but for `Double` `NaN.equals(NaN) == true`

tpietzsch:
Well, that begs the question then: Should imglib2 `DoubleType` then compare like a `double` or like a `Double`... Aaaaaarghhhhhhh.

We could even go for:
`DoubleType.equals()` compares like `Double`
`DoubleType.valueEquals()` like `double`.
It seems intuitively wrong, but is it? After all, the "value" of a `DoubleType` is a `double`.
It is an edge case, but would certainly be good to define and document.

tpietzsch:
For completeness:

```java
double d1 = Double.NaN;
double d2 = Double.NaN;
Double D1 = Double.valueOf( d1 );
Double D2 = Double.valueOf( d2 );

System.out.println( "(d1 == d2) = " + ( d1 == d2 ) );
System.out.println( "D1.equals( D2 ) = " + D1.equals( D2 ) );
System.out.println( "D1.compareTo( D2 ) = " + D1.compareTo( D2 ) );
System.out.println( "Double.compare( d1, d2 ) = " + Double.compare( d1, d2 ) );
System.out.println( "Comparator.comparingDouble( Double::doubleValue ).compare( D1, D2 ) = " + Comparator.comparingDouble( Double::doubleValue ).compare( D1, D2 ) );
```

prints
```
(d1 == d2) = false
D1.equals( D2 ) = true
D1.compareTo( D2 ) = 0
Double.compare( d1, d2 ) = 0
Comparator.comparingDouble( Double::doubleValue ).compare( D1, D2 ) = 0
```

tpietzsch:
Starting from this PR, I tried to clean up implementations of `equals`, `valueEquals`, `hashCode`, and `compareTo` in all existing Types (https://github.com/maarzt/imglib2/pull/1).

In general the idea was:
* Make `valueEquals` an equivalence relation (in particular, `o.valueEquals(o)==true` always)
* Make `equals` and `valueEquals` always consistent.
* For comparisons and hashcodes fall back to `java.lang` classes if possible (e.g., https://github.com/imglib/imglib2/commit/e981814d8318c3820c35907f67b42a998653d916, https://github.com/imglib/imglib2/commit/b8addb1fd902fc930e7a64a315b96dad9f0deda3)

I hit some questions that should be answered before going on:
This is going in the direction where "`equals()`" is always "type-check + `valueEquals()`". That is fine, but it is easy for inconsistencies to creep in, especially, with `equals`, `valueEquals`, and `hashCode` implementations strewn through the type hierarchy in abstract classes.
Concrete example: `AbstractIntegerType` implements `equals` to call `valueEquals`, and implements `valueEquals` and `hashCode` to fall back on `long` values. It is now easy for a subclass to override `valueEquals` and thereby cause `equals` and `hashCode` to become inconsistent.

There is no perfect solution for this scenario.
I would suggest to just never call `valueEquals` from `equals` in non-leafs of the type hierarchy.
That is, this https://github.com/imglib/imglib2/blob/b8addb1fd902fc930e7a64a315b96dad9f0deda3/src/main/java/net/imglib2/util/Util.java#L983-L996 should be removed. It could be replaced by
```
	public static < T extends ValueEquals< T > > boolean valueEqualsObject( T a, Object b )
	{
		if ( !a.getClass().isInstance( b ) )
			return false;
		return a.valueEquals( ( T ) b );
	}
```
which cannot be called from intermediate abstract `Types`. (I would rather remove it do it explicitly in the equals methods like https://github.com/imglib/imglib2/blob/b8addb1fd902fc930e7a64a315b96dad9f0deda3/src/main/java/net/imglib2/type/numeric/complex/ComplexFloatType.java#L300-L305
because that's not really longer and one less indirection)

Opinions?
If there are no objections, then I would ask that somebody continues with cleaning up the remaining `Type`s. (@maarzt could you do it?)

axtimwalde:
In, the `AbstractIntegerType.equals` implementation, you could call `AbstractIntegerType.this.valueEquals` to avoid the unwanted default virtualization?

maarzt:
@tpietzsch I updated the PR. I walked through all `Types` and updated hashCode, equals, valueEquals and compareTo wherever I cloud.

Don't miss the fancy `NumericTypeTest`. It tests the basic functionality of every numeric type. Just to make it harder to break things.

maarzt:
I think we should squash the commit

src/main/java/net/imglib2/type/numeric/complex/AbstractComplexType.java
```diff
@@ -189,9 +190,9 @@ public boolean equals( final Object obj )
 	@Override
 	public int hashCode()
 	{
-		final int hash1 = Double.hashCode( getRealDouble() );
-		final int hash2 = Double.hashCode( getImaginaryDouble() );
-		return Util.combineHash( hash1, hash2 );
+		final int iHash = Double.hashCode( getRealDouble() );
+		final int rHash = Double.hashCode( getImaginaryDouble() );
```
imagejan: You probably want to swap these, so that `rHash <= getRealDouble()` and `iHash <= getImaginaryDouble()`, no?

maarzt:
@tpietzsch Ok it's ready from my side. I looked on all types, and updated equals, valueEquals, hashCode and compareTo methods. Most notably the behavoir of `FloatType#equals` and `Double#equals` as well as valueEquals, behaves like the equals method of the boxed Java types.

maarzt:
@tpietzsch It's again ready for review & maybe merge.

src/main/java/net/imglib2/type/numeric/complex/AbstractComplexType.java
maarzt: done

maarzt:
@tpietzsch I squashed it, should be ready for merge ;)
I merged your changes into this PR, squashed and rewrote commit messages.
To source code didn't change with one exception:
1. In the JUnit test `NumericTypeTest`: I changed the tests for the method `compareTo`. But that's rather boring.

tpietzsch:
Thanks @maarzt and contributors to the discussion! I think this is a nice step forward in consistency.

PR #227, opened 2018-08-30
Suggestion to introduce method Casts.uncecked()

maarzt:
I often dreamed of  having this method Casts.uncecked(...). It makes it easy to write an unchecked cast.
No need to add two lines of code + a @SuppressWarnings annotation.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
 	</parent>
 
 	<artifactId>imglib2</artifactId>
-	<version>5.4.1-SNAPSHOT</version>
+	<version>5.5.1-SNAPSHOT</version>
 
 	<name>ImgLib2 Core Library</name>
 	<description>A multidimensional, type-agnostic image processing library.</description>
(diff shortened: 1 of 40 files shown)


maarzt:
This PR was accidentally created here.
(There seems to be a github, that creates the PR in the main repo, although a specified it to be a PR between two branches in my fork)

tpietzsch:
Looks really useful, though...

ctrueden:
This function is not really an ImgLib2-specific thing. I think it belongs in SciJava.

See also
https://github.com/scijava/scijava-common/blob/scijava-common-2.75.0/src/main/java/org/scijava/util/Types.java#L753-L762

We could also add a cast without specifying the T, and/or a `Nil<T>` cast to ensure type safety including generics. This portion of scijava-common is slated to become `scijava-types` or even `scijava-util`, to minimize footprint. Although we will need to find a balance between footprints of "number of dependencies" vs. "size of each dependency".

tpietzsch:
I think a simple isolated utility method like this can be replicated in both projects without any harm.
Downstream users can choose freely to use whichever they want interchangeably.
It doesn't justify adding scijava as a imglib2 dependency nor otherwise.
(Not saying that there are no things that would justify a dependency, but this is not one of them, IMO)

PR #228, opened 2018-09-03
Add StopWatch utility from imglib2-cache

gab1one:
This fixes a split package problem caused by this class living in imglib2-cache. See also: https://github.com/imglib/imglib2-cache/pull/6

### DIFF
--- a/src/main/java/net/imglib2/util/StopWatch.java
+++ b/src/main/java/net/imglib2/util/StopWatch.java
@@ -33,6 +33,8 @@
  */
 package net.imglib2.util;
 
+import java.text.DecimalFormat;
+
 /**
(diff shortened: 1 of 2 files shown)


maarzt:
I added some methods to make life easier. Be amazed by the before, after comparison:

Before (to difficult)
```java
StopWatch sw = new StopWatch();
sw.start();
methodToMessureExecutionTimeOf();
long time = sw.nanoTime;
System.out.println( time / 1000000.0 + " ms" );
```

After (easy)
```java
StopWatch sw = StopWatch.createAndStart();
methodToMessureExecutionTimeOf();
System.out.println( sw );
```

PR #229, opened 2018-10-18
LinearRangeConverter: to map a value into a [min, max] display range.

acardona:
A Converter to enable viewing an image with its values mapped to a [min, max] range.
Similar to ImageJ's min-max display range, but non-destructive.

### DIFF
--- a/src/main/java/net/imglib2/converter/LinearRangeConverter.java
+++ b/src/main/java/net/imglib2/converter/LinearRangeConverter.java
@@ -1,0 +1,83 @@
+/*
+ * #%L
+ * ImgLib2: a general-purpose, multidimensional image processing library.
+ * %%
+ * Copyright (C) 2009 - 2018 Tobias Pietzsch, Stephan Preibisch, Stephan Saalfeld,
+ * John Bogovic, Albert Cardona, Barry DeZonia, Christian Dietz, Jan Funke,
+ * Aivar Grislis, Jonathan Hale, Grant Harris, Stefan Helfrich, Mark Hiner,
+ * Martin Horn, Steffen Jaensch, Lee Kamentsky, Larry Lindsey, Melissa Linkert,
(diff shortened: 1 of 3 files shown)


axtimwalde:
Sure you cannot live with https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/converter/RealUnsignedByteConverter.java and https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/converter/RealARGBConverter.java ?  Both implement https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/display/AbstractLinearRange.java which looks like significant code duplication with this pull request.

acardona:
Thanks, I didn't know about those. Didn't occur to me to look into the display package.

To remark that this pull request provides a generic type-based implementation that is currently missing. Applies to any subclasses of RealType with Type-based math. (Has to be RealType and not NumericType because of the Comparator methods.)

And the existing concrete classes do not work for e.g. UnsignedShortType, which is the type of the images I was going to use the new classes for.

What do you propose? My code is more general. Surely also less performant, but I haven't tested performance yet.

axtimwalde:
`LinearRange` and `AbstractLinearRange` apply to all combinations of `RealType` so I think they should be extended at least. Sure about `UnsignedShortType` as a target of the mapping?  For display purposes as you mention in the motivation, you would always want the target to be either `ARGBType` or `UnsignedByte` (except on display devices that I do not have access to).  Do you need this in your scripting framework?  In Java, I always write simple converters like this as lambdas which is shorter than construction plus use of an existing class.  Just wondering where to put this best, if it has an actual use case.

axtimwalde:
No new commits addressing the above questions.  Is this still relevant or ok to close?

acardona:
Thanks, just close it. Will reopen if I get back to it. Found other ways to get the operation I wanted run fast.

PR #230, opened 2018-10-18
ImgUtil methods

acardona:
Added two missing methods to ImgUtil:

1. to copy an Img into a short[]
2. to copy an Img into another Img of the same type.


axtimwalde:
Thanks!

PR #231, opened 2018-12-28
Add ConstantUtils.constantRandomAccessibleInterval without explicitly passing number of dimensions

hanslovsky:
The explicitly passed parameter numDimensions is not required as this can be deducted from the provided interval.


tpietzsch:
Should we deprecate the other signature? I don't really see a use case.

hanslovsky:
Done in 83a5538

Added some simple tests as well.

tpietzsch:
👍 thanks!

PR #234, opened 2019-01-20
ClassCopier: Use the classloader of the class that is copied

gab1one:
Previously the system classloader was used as parent of the internal classloader, which can create problems in environments with multiple classloaders (e.g. OSGi).

### DIFF
--- a/src/main/java/net/imglib2/loops/ClassCopier.java
+++ b/src/main/java/net/imglib2/loops/ClassCopier.java
@@ -67,7 +67,7 @@ class ClassCopier< T >
 	public Class< ? extends T > copy()
 	{
 		@SuppressWarnings( "unchecked" )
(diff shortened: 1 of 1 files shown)


maarzt:
Great! But we still need to test if this fixes the problem with KNIME.
I will check if it works with imglib2-algorithms.

maarzt:
@gab1one Can you test if this fixes the problem in KNIME Image Processing?

gab1one:
Yes it seems to do that.

tpietzsch:
@hanslovsky Would you be able to test whether this solves the Spark problem too?

hanslovsky:
Yep, fixes it. I will test `Gauss3` from imglib2-algorithm as well.

tpietzsch:
Thank you guys!

hanslovsky:
`Gauss3` is fixed with this, as well. This is my spark test project:
https://github.com/hanslovsky/spark-test-imglib2-loopbuilder

PR #236, opened 2019-01-23
Partition.java does not check the validity of input iterators

emopers:
`Partition.java` calls `j.previous()` on `java.util.Iterator j` and `i.next()` on `java.util.Iterator i` 
without checking if there are any elements to iterate over. Because the method is public and the iterators are obtained from inputs, they could be invalid (e.g., an empty list) and lead to an exception. This pull request adds a `hasNext()` and a `hasPrevious()` check.

This violation is present in other parts of this class as well. If this fix is acceptable to you, we can
submit another pull request with fixes for all other cases.

### DIFF
--- a/src/main/java/net/imglib2/util/Partition.java
+++ b/src/main/java/net/imglib2/util/Partition.java
@@ -37,6 +37,7 @@ package net.imglib2.util;
 import java.util.Comparator;
 import java.util.List;
 import java.util.ListIterator;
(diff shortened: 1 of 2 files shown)


ctrueden:
Looks good to me! What do you think, @tpietzsch?

ctrueden:
I retract my LGTM; see https://github.com/imglib/imglib2/pull/259#issuecomment-521686798.

PR #237, opened 2019-01-23
ImgUtil.copy RandomAccessibleInterval to an IterableInterval, multithreaded

acardona:
New utility method ImgUtil.copy(RandomAccessibleInterval, IterableInterval, int) for multi-threaded copying with `int` threads.

Addresses the issue of copying "images" that are actually transformed views using linear interpolators and others, which can be costly and, from a scripting framework, it is not entirely trivial to partition a RandomAccessibleInterval for multi-threaded processing.
While it could be done via Views.interval, it is simpler to use a utility method to do it all.

### DIFF
--- a/src/main/java/net/imglib2/util/ImgUtil.java
+++ b/src/main/java/net/imglib2/util/ImgUtil.java
@@ -35,8 +35,9 @@
 package net.imglib2.util;
 
 import net.imglib2.Cursor;
-import net.imglib2.RandomAccess;
+import net.imglib2.RandomAccessibleInterval;
 import net.imglib2.img.Img;
+import net.imglib2.loops.LoopBuilder;
 import net.imglib2.type.BooleanType;
 import net.imglib2.type.Type;
 import net.imglib2.type.numeric.IntegerType;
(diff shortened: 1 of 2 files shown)


maarzt:
The copy method in this PR is rather complex, because it needs to deal with all the details of multi-threading. This motivated me to add multi-threading to `LoopBuilder` https://github.com/imglib/imglib2/pull/250. The multi-threaded `LoopBuilder` allows to dramatically simplify the copy method:

```java
public static < T extends Type< T >> void copy(
	final RandomAccessibleInterval< T > src,
	final RandomAccessibleInterval< T > dest )
{
	LoopBuilder.setImages( src, dest ).multiThreaded().forEachPixel(
		(s, d) -> d.set( s )
	);
}
```

BUT: The destination's type is now an `RandomAccessibleInterval` rather than an `IterableInterval`. 

@acardona I would like to know your opinion: Is there a benefit of choosing `IterableInterval` as destination type? What's the motivation for having `n_threads` as a parameter?

acardona:
Hi @maarzt,
1. ```IterableInterval``` is preferable if the destination type was ```IterableInterval```. Otherwise wrapping is needed, which is very undesirable.
2. ```n_threads``` is needed when running e.g. 4 only threads doing an expensive operation (due to memory constraints) but then the copying can fully use e.g. 32/4 = 8 threads for the copying of each result image into e.g. an ```ArrayImg```. (This is pulled from a real world situation).

The ```ImgUtil.copy``` serves a real need in scripting languages that can't optimize loops like for example jython.

ctrueden:
The pyimagej project would greatly benefit from a performant single-line RAI-to-RAI (and/or RAI-to-II) copy method, which does not require writing any lambdas.

@maarzt Does a way currently exist to achieve such a performant single-line copy operation currently? If not, can we get something added to the library for this soon? We will need it for the pyimagej 1.0.0 release.

ctrueden:
Thanks, @maarzt!

PR #238, opened 2019-01-24
Extract MixedTransform creation and Interval manipulation methods from Views

chaubold:
This PR basically moves a lot of the code hidden in the static methods in `Views` into helper methods that create `MixedTransform`s or modify `Interval`s. 

The intention behind this refactoring is to simplify the application of a transformation -- e.g. one that was applied to image data -- also to other corresponding objects like metadata.

The API of `Views` remains unchanged, and the tests that currently pass on master also pass after this refactoring.

CC: @gab1one @MarcelWiedenmann @dietzc @HedgehogCode

### DIFF
--- a/src/test/java/net/imglib2/util/IntervalsTest.java
+++ b/src/test/java/net/imglib2/util/IntervalsTest.java
@@ -1,0 +1,133 @@
+/*
+ * #%L
+ * ImgLib2: a general-purpose, multidimensional image processing library.
+ * %%
+ * Copyright (C) 2009 - 2018 Tobias Pietzsch, Stephan Preibisch, Stephan Saalfeld,
+ * John Bogovic, Albert Cardona, Barry DeZonia, Christian Dietz, Jan Funke,
+ * Aivar Grislis, Jonathan Hale, Grant Harris, Stefan Helfrich, Mark Hiner,
+ * Martin Horn, Steffen Jaensch, Lee Kamentsky, Larry Lindsey, Melissa Linkert,
+ * Mark Longair, Brian Northan, Nick Perry, Curtis Rueden, Johannes Schindelin,
+ * Jean-Yves Tinevez and Michael Zinsmaier.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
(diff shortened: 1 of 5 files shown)


maarzt:
When reviewing this, I was very surprised by the existing of the `Views.offset` method. My confusion looked like this:
* Does it do the same like `Views.translate`
* Did maybe somebody accidentally added this method, being not aware that `Views.translate` exists.
* Then I noticed the sign is opposite to `Views.translate`
* I looked up the history of this method.

To me it's obvious, this method is there for historical reasons, and need's be deprecated. (This would be to avoid future confusing, bugs & resulting debug sessions). We should not add new confusing methods, Intervals.offset & MixedTransform.offset (it would be to much code duplication anyway).

maarzt:
Very nice PR. I see that the new methods in `Intervals` are useful.

@chaubold Please add Unit Tests, to ensure that the PR is correct.

chaubold:
Thanks for the review @maarzt! I will add a few unit tests directly for the Intervals and MixedTransforms methods, but they are already covered quite well by the existing tests for Views.

And indeed, offset is equal to translating by the negative. I don’t think I have enough imglib2 streed creds to be able to decide whether that should be deprecated or not, I would leave that up to you and @tpietzsch

maarzt:
Yes, deprecating the ```offset``` method could be a separate PR. But still I'm opposed to add new methods that are named ```offset``` in this method.

To explain why: The intention off ```Views.offset``` is to remove an offset. It should be therefor name ```Views.removeOffset```. The name ```Views.offset``` sounds like it adds an offset, but that's the opposite of what it does. It's therefor really confusingly named.

tpietzsch:
> The intention off `Views.offset` is to remove an offset.

Naming is tricky here... If you add `t` to the source image coordinates (`x_view = x_source + t`), then you are shifting the image by `-t`. So it depends on what you are talking about. Are you offsetting coordinates or the image? (Or, another way to look at it, are you looking at it taking the source coordinate system or the view coordinate system as reference.)

Anyway, you are right. `offset` should be deprecated. It doesn't make sense to have both `offset` and `translate`.

maarzt:
Ok, I deprecated Views.offset on master. (I got green light from @tpietzsch).

@chaubold A unit test should be as simple as possible. Many existing tests are overly complicated. For example, a good simple unit test for Views.rotate looks like this:
```java
package net.imglib2.util;

import net.imglib2.Interval;
import net.imglib2.test.ImgLib2Assert;
import org.junit.Test;

public class IntervalsTest
{

	@Test
	public void testRotate() {
		Interval input = Intervals.createMinMax( 1, 2, 3, 5, 7, 9 );
		Interval result = Intervals.rotate( input, 1, 0 );
		Interval expected = Intervals.createMinMax( 2, -5, 3, 7, -1, 9 );
		ImgLib2Assert.assertIntervalEquals( expected, result );
	}
}
```

hanslovsky:
> Anyway, you are right. offset should be deprecated. It doesn't make sense to have both offset and translate.

I disagree, I do like to be able to use `Views.offset`, when I would have to invert the array first before passing it to `Views.translate`. Unless there is a `Views.translateInverse` or similar, of course.

src/test/java/net/imglib2/util/IntervalsTest.java
```diff
@@ -0,0 +1,133 @@
...
+        final Interval expected = Intervals.createMinMax( 11, 2, 3, 15, 7, 9 );
+        ImgLib2Assert.assertIntervalEquals( expected, result );
+    }
+
+    @Test
+    public void testTranslateUsesCorrectOverload() {
+        final Interval input = Intervals.createMinMax( 1, 2, 5, 7 );
+        // when the number of axes equals 2, then both translate() methods are applicable. 
+        // Which one is used?
+        final Interval result = Intervals.translate( input, 10, (long)1 );
```
chaubold: I've got one more question for you: `Intervals` has a `translate(interval, dimension, shiftAtDimension)` method already, and I added `translate(Interval interval, long... translation)`. When the interval is 2D, this means the number of arguments to both methods is the same, the only difference being `int` vs `long` for the 2nd argument.

I used `translate(Interval interval, long... translation)` because that matches the API of `Views.translate(RandomAccessibleInterval input, long... translation)`. But now that I noticed that signature clash I am wondering whether we should adjust the naming of one of the two methods... I'd prefer to rename the translation method that only translates a single dimension to something like `translateDim`, but that would break the current `Intervals` API.

Any opinions @maarzt @tpietzsch ?

chaubold: I just checked: inside imglib2 the existing `Intervals.translate(interval, dim, shift)` method is only used inside the `LoopPerformanceBenchmark`, but that doesn't mean nobody else is using it...

maarzt: I used the ```Intervals.translate(interval, dim, shift)``` in imglib2-algorithms. So it's very likely used in other places as well. It's terribly sad, we **must not rename** it. Because then the other translate method would chip in. And the method's behavior will be slightly different. It will introduce bugs, which are impossible to find and hard to notice.

The solution is to: change the signature ```Views.translate(RandomAccessibleInterval input, long... translation)``` to ```Views.translate(RandomAccessibleInterval input, long[] translation)```.

@chaubold In this PR please use ```Intervals.translate(Interval input, long[] translation)```. I will make a separate PR to change ```Views.translate```.

maarzt: Sorry I was wrong with the previous message there is no ```Views.translate(image, shift, dim)```

maarzt: @chaubold We don't need to fix this problem in this PR. Just create an issue.

maarzt: I created the issue

tpietzsch:
I applied imglib2 code style and merged.

PR #245, opened 2019-02-07
Add Views.translateInverse() to replace Views.offset()

tpietzsch:
This adds `Views.translateInverse` to replace the recently deprecated (because of naming) `Views.offset`. Following recently merged (#238) split of functionality out of `Views`, corresponding `Intervals.translateInverse` and `MixedTransforms.translateInverse` methods are also introduced (and used).

@hanslovsky @maarzt Does this work for everybody?


hanslovsky:
Thanks @tpietzsch  works for me, I like the name better than the old one.

PR #246, opened 2019-02-12
Add convenience methods RandomAccess.get( position )

hanslovsky:
Shortcut to
```java
setPosition( position );
get();
```

This is super useful in places where efficiency is not of the essence, e.g. labeling seeds for watersheds (pseudo-code):
```
RandomAccess access;
List<Point> seedPoints;
seedPoints.forEach(p -> access.get(p).set(index.incrementAndGet());
```

### DIFF
--- a/src/main/java/net/imglib2/RandomAccess.java
+++ b/src/main/java/net/imglib2/RandomAccess.java
@@ -73,7 +73,7 @@ public interface RandomAccess< T > extends Localizable, Positionable, Sampler< T
 	 * @param position
 	 * @return value of the the {@code RandomAccess} at {@code position}.
 	 */
-	default T get( long... position )
+	default T getAt( long... position )
 	{
 		assert position.length == numDimensions();
 
@@ -99,7 +99,7 @@ public interface RandomAccess< T > extends Localizable, Positionable, Sampler< T
 	 * @param position
 	 * @return value of the the {@code RandomAccess} at {@code position}.
 	 */
-	default T get( int... position )
+	default T getAt( int... position )
 	{
 		assert position.length == numDimensions();
 
@@ -125,7 +125,7 @@ public interface RandomAccess< T > extends Localizable, Positionable, Sampler< T
 	 * @param position
 	 * @return value of the the {@code RandomAccess} at {@code position}.
 	 */
-	default T get( Localizable position )
+	default T getAt( Localizable position )
 	{
 		assert position.numDimensions() == numDimensions();
 
--- a/src/main/java/net/imglib2/RandomAccess.java
+++ b/src/main/java/net/imglib2/RandomAccess.java
@@ -73,7 +73,7 @@ public interface RandomAccess< T > extends Localizable, Positionable, Sampler< T
 	 * @param position
 	 * @return value of the the {@code RandomAccess} at {@code position}.
 	 */
-	default T getAt( long... position )
+	default T setPositionAndGet(long... position )
 	{
 		assert position.length == numDimensions();
 
@@ -99,7 +99,7 @@ public interface RandomAccess< T > extends Localizable, Positionable, Sampler< T
 	 * @param position
 	 * @return value of the the {@code RandomAccess} at {@code position}.
 	 */
-	default T getAt( int... position )
+	default T setPositionAndGet(int... position )
 	{
 		assert position.length == numDimensions();
 
@@ -125,7 +125,7 @@ public interface RandomAccess< T > extends Localizable, Positionable, Sampler< T
 	 * @param position
 	 * @return value of the the {@code RandomAccess} at {@code position}.
 	 */
-	default T getAt( Localizable position )
+	default T setPositionAndGet(Localizable position )
 	{
 		assert position.numDimensions() == numDimensions();
 
--- a/src/main/java/net/imglib2/RandomAccess.java
+++ b/src/main/java/net/imglib2/RandomAccess.java
@@ -73,7 +73,7 @@ public interface RandomAccess< T > extends Localizable, Positionable, Sampler< T
 	 * @param position
 	 * @return value of the the {@code RandomAccess} at {@code position}.
 	 */
-	default T setPositionAndGet(long... position )
+	default T setPositionAndGet( long... position )
 	{
 		assert position.length == numDimensions();
 
@@ -99,7 +99,7 @@ public interface RandomAccess< T > extends Localizable, Positionable, Sampler< T
 	 * @param position
 	 * @return value of the the {@code RandomAccess} at {@code position}.
 	 */
-	default T setPositionAndGet(int... position )
+	default T setPositionAndGet( int... position )
 	{
 		assert position.length == numDimensions();
 
@@ -125,7 +125,7 @@ public interface RandomAccess< T > extends Localizable, Positionable, Sampler< T
 	 * @param position
 	 * @return value of the the {@code RandomAccess} at {@code position}.
 	 */
-	default T setPositionAndGet(Localizable position )
+	default T setPositionAndGet( Localizable position )
 	{
 		assert position.numDimensions() == numDimensions();
 
--- a/src/main/java/net/imglib2/RandomAccess.java
+++ b/src/main/java/net/imglib2/RandomAccess.java
@@ -40,6 +40,7 @@ package net.imglib2;
  * @author Tobias Pietzsch
  * @author Stephan Preibisch
  * @author Stephan Saalfeld
+ * @author Philipp Hanslovsky
  */
 public interface RandomAccess< T > extends Localizable, Positionable, Sampler< T >
 {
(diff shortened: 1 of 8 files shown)


maarzt:
I see how it's useful. Reading a pixel from a RandomAccessible is hard, it requires three steps:
```java
RandomAccess access = rai.randomAccess();
access.setPosition(position);
value = access.get();
```
Which could be reduced to just one line:
```java
value = rai.randomAccess().get(position);
```

But I see some problems with this approach:
* As it's now the interface ```RandomAccess``` clearly shows, that one needs one RandomAccess per thread. The ```get(position)``` invites concurrent usage. Which, if misused, introduces subtle bugs.
* It's kind of surprising that ```get(position)``` changes the position. It should be renamed ```setPositionAndGet(position)```, because that would exactly describe what the method does.
* We would need to have multiple methods like this. One for ```long[]``` and one for ```Localizable``` and maybe one for ```int[]```. That clutters the interface of RandomAccess a lot.

I like the interface ```RandomAccess``` as it is. I think it's clear and easy to understand. In my opinion the new method would decrease readability. Which outwaits the advantages.

Regarding the motivating example: A method ```forEachPoint(randomAcces, points, action)``` might be the best solution? Or maybe a nice method that allows to perform a action for a region of interest, combined with an imglib2 point region of interest? (Disclaimer I don't no nothing about imglib2 rois)

hanslovsky:
> As it's now the interface `RandomAccess` clearly shows, that one needs one RandomAccess per thread. The `get(position)` invites concurrent usage. Which, if misused, introduces subtle bugs.

I do not see how code would be more prone to these kind of bugs with the suggested changes. After all, the `get(position)` is not a member of the `RandomAccessible`, but of the `RandomAccess` that should be created separately for each thread anyway.

> It's kind of surprising that `get(position)` changes the position. It should be renamed `setPositionAndGet(position)`, because that would exactly describe what the method does.

Good point, I do prefer something like `getAt(position)`, though, as this is very precise about the intent, i.e. get the value at the specified position.

> We would need to have multiple methods like this. One for `long[]` and one for `Localizable` and maybe one for `int[]`. That clutters the interface of RandomAccess a lot.

The methods are already present in this PR and I do not think they add too much clutter. Especially as there are already 24 methods in the interface, counting inherited methods.

tpietzsch:
I think @maarzt has a good point, that it is unexpected that the position is changed. So I would not merge this PR.

How about making a separate class/interface that wraps a `RandomAccess` and provides the `get(position)` methods? Then for convenience we could discuss adding a default method to `RandomAccessible` to create one of these wrappers. Unfortunately the name `RandomAccess` is already taken, because it would be a good name for the new thing...

hanslovsky:
> I think @maarzt has a good point, that it is unexpected that the position is changed. So I would not merge this PR.

What if I go with @maarzt's suggestion for a more verbose name? I still do like the concise name better, especially as the doc is very specific, but if the general opinion prefers a verbose name I am willing to change it that way.

If only this was `c++` and we had a `const` keyword, then this would be a non-issue (`const` methods do not change state of object).

> How about making a separate class/interface that wraps a `RandomAccess` and provides the `get(position)` methods? Then for convenience we could discuss adding a default method to `RandomAccessible` to create one of these wrappers. Unfortunately the name `RandomAccess` is already taken, because it would be a good name for the new thing...

Is this really necessary, even if we used a more verbose name?
If so, I suggest an interface (pseudo-code):
```
interface DirectAccess<T> {
    // DirectAccess does not know if implementations use setPosition, so name should not reflect that
    T getAt(long... position);
    T getAt(int... position);
    T getAt(Localizable position);

    static <U> DirectAccess<U> from(RandomAccess<U> access) {
        return new FromRandomAccess(access);
    }
    
    static class FromRandomAccess<U> implements DirectAccess<T> {
          // add methods from pull request here
    }
}

**Comment by tpietzsch** [comment]
> What if I go with @maarzt's suggestion for a more verbose name?
I could live with that, but I see advantages to separate this functionality into a `DirectAccess` interface, though. These could be implemented to not need a `RandomAccess` at all. For example, `ArrayImg` (and probably others) could even implement a `DirectAccess` that is stateless, so `ArrayImg.directAccess()` could always return the same instance.

**Comment by hanslovsky** [comment]
Fair point, I will add those changes later. I do not want to be too eager with another commit so we can think about the interface and naming a little more.

**Comment by tpietzsch** [comment]
> For example, `ArrayImg` (and probably others) could even implement a `DirectAccess` that is stateless, so `ArrayImg.directAccess()` could always return the same instance.

Argh, damnit. It cannot be stateless because it has to have a `Type` instance.
Ok, then my point is probably moot... Sorry, I think  I'm too tired today...

**Comment by hanslovsky** [comment]
For whoever is using kotlin (is there anyone besides me and @skalarproduktraum?), you can just use kotlin extension functions on the random access (or even `RandomAccessible`) to get this behaviour. Just add (and modify as needed) these lines, wherever appropriate:
```kotlin
operator fun <T> RandomAccess<T>.get(vararg pos: Long): T {
    this.setPosition(pos)
    return get()
}

operator fun <T> RandomAccessible<T>.get(vararg pos: Long): T {
    return randomAccess().get(*pos)
}
```

You can then run:
```kotlin
val a: RandomAccessible<T> = ....
a.randomAccess()[1, 2, 3]
a[1, 2, 3]
```

maarzt:
@hanslovsky Renaming the method from ```get``` to ```getAt``` doesn't solve the disadvantages in this PR. Let me try to give another example to show what's problematic. The PR creates the following pitfall:
```java
RandomAccess<T> ra = image.randomAccess();
T pixelA = ra.getAt(1,2,3);
T pixelB = ra.getAt(4,5,6);
```
**The problem with this example:** ```pixelA == pixelB```
There are many more variations of this pitfall.

Your Kotlin example suggest another solution:
> You can then run:
> ```java
> val a: RandomAccessible<T> = ....
> a.randomAccess()[1, 2, 3] // UGLY
> a[1, 2, 3] // BEAUTIFUL
> ```
I would be better to instead add the ```T get(long... position)``` method to the  ```RandomAccessible``` interface. @hanslovsky How do you think about that?

hanslovsky:
> I would be better to instead add the T get(long... position) method to the RandomAccessible interface. @hanslovsky How do you think about that?

@maarzt I agree, that it makes more sense to add it to the `RandomAccessible` interface. Unfortunately, I am rather busy right now, but I will be in Dresden for the Learnathon in June, so we could talk about it there (also pinging @tpietzsch)

tpietzsch:
@hanslovsky @maarzt Ok, let's talk about it at the Learnathon

ctrueden:
Adding a direct value getter to the `RandomAccessible` interface would be very convenient. Perhaps too convenient. My understanding is that we intentionally did not have this in the past, for performance reasons? But if @tpietzsch says it would be OK, then I'm all for it! :+1:

Also, just a heads up that I wrote a thing sorta like that during the Janelia 2016 hackathon, which was never completed or merged (although it is functional). See these commits:

* https://github.com/imagej/imagej-common/commit/f467470585ede024450a6dfa911d9893819398f2
* https://github.com/imagej/imagej-common/commit/b0603cef203710844d507216bc075147dbd3a474

hanslovsky:
@maarzt @tpietzsch What do you think? This now incoropartes the changes that we agreed upon in Dreden.

src/main/java/net/imglib2/RandomAccess.java
```diff
@@ -53,4 +54,116 @@
...
+	 *
+	 * // Correct:
+	 * // Use individual RandomAccesses to query a and b
+	 * a = randomAccess_A.setPositionAndGet( positionA );
+	 * b = randomAccess_B.setPositionAndGet( positionB ); // this is fine because a different RandomAccess is used
+	 * difference = a.getRealDouble() - b.getRealDouble();
+	 * }
+	 * </pre>
+	 */
+	default T setPositionAndGet( long... position )
```
maarzt: @hanslovsky I didn't like the javadoc like it was before. It said, that this method has performance issues, that's just not the case. But at the same time I was missing the Warning, that you can see now, which I think is very important.

src/test/java/net/imglib2/RandomAccessTest.java
```diff
@@ -0,0 +1,64 @@
...
+	@Test
+	public void testSetPositionAndGet()
+	{
+		// setup
+		final Img< IntType > image = ArrayImgs.ints( new int[]{ 1, 2, 3, 4, 5, 6 }, 3, 2 );
+		RandomAccess< IntType > randomAccess = image.randomAccess();
+		// process & test
+		assertEquals( new IntType( 4 ), randomAccess.setPositionAndGet( new Point( 0L, 1L ) ) );
+		assertEquals( new IntType( 5 ), randomAccess.setPositionAndGet( 1L, 1L ) );
+		assertEquals( new IntType( 3 ), randomAccess.setPositionAndGet( 2, 0 ) );
```
maarzt: @hanslovsky :tada: Hey look are need and simple this test is now! Much easier to read, understand and debug, and still does the job amazingly well.

src/main/java/net/imglib2/RandomAccess.java
maarzt: Your are right `ListRandomAccess` is an exception. And also that `setPositionAndGet(...)` is just a shortcut, and exactly the same as two consecutive calls to `setPosition` and `get`. But still, I have a strong gut feeling that it will invite new imglib2 users to use it wrongly, and that the warning will be very useful.

PR #247, opened 2019-02-12
Rename MixedTransform to InverseTransform

maarzt:
The class ```MixedTransforms``` needs to be renamed. Because it's currently strange that:
```MixedTransforms.translate(...)``` actually return an inverse transformation. ```MixedTransforms``` does behave like that way, because ```Views``` always use the inverse transformation. But reflect this intention and behavior, we should rename the recently added class from ```MixedTransforms``` to ```ViewTransforms```.


maarzt:
I still need to update the javadocs.

PR #249, opened 2019-03-05
Simple Conversion between RealTypes

maarzt:
When I need to convert between two RealTypes a usally us a lambda expression like this:
```java
Converter<RealType, RealType> converter = (i, o) -> o.setReal(i.getRealDouble());
```
But this converter is not optimal in term of performance (To convert ByteType -> IntType you would not need do cast to double), a precision (UnsignedLongType can be poorly represented as double).

This PR adds three methods to fix this:

* Getting converter between two RealTypes, which is optimal in terms of performance and precision, is as easy as:
```java
Converter<UnsignedByteType, LongType> converter = 
     RealTypeConverter.getConverter(new UnsignedByteType(), new LongType());
```
* Converting the type of a RandomAccessibleInterval:
```java
RandomAccessibleInterval<LongType> converter = 
     RealTypeConverter.convert(image, new LongType());
```
* Copy pixels between to RandomAccessibleIntervals:
```java
CopyPixels.fromTo(imageA, imageB);
```


maarzt:
I'm not sure about the names of the static methods, or to which class the should belong: ```RealConverters.getConverter```, ```RealConverters.convert``` and especially ```CopyPixels.fromTo```.

And I think about adding more functionality: Converters that bound the value to minimum and maximum. And converters allow to scale the values.

tpietzsch:
@StephanPreibisch @axtimwalde I just noticed (because this PR touches it), that [`Util.round()`](https://github.com/imglib/imglib2/blob/fad95aa5b48398029227cb676e3a44b78e79bc4c/src/main/java/net/imglib2/util/Util.java#L501-L509) behaviour differs from `Math.round()` for negative numbers. E.g., `Math.round()` rounds `-0.5` to `0`, while `Util.round()` rounds `-0.5` to `-1`.

Do you remember what the reason was to do it like this?
And if there was no specific reason, should we change it to just do `Math.round()`?

src/main/java/net/imglib2/converter/RealTypeConverters.java
```diff
@@ -0,0 +1,249 @@
...
+			this.provider = new ClassCopyProvider<>( converterClass, Converter.class );
+		}
+
+		public Converter create( RealType< ? > in, RealType< ? > out )
+		{
+			return provider.newInstanceForKey( Arrays.asList( in.getClass(), out.getClass() ) );
+		}
+	}
+
+	public static class DoubleConverter implements Converter< RealType< ? >, RealType< ? > >
```
tpietzsch: With these classes being `public` could you add javadoc, please.
Can be very minimal, e.g.,
```java
/**
 * Converts via {@link RealType#getRealDouble}.
 */
```
 is better than nothing.

tpietzsch:
> A general copy-pixels method would be nice though.
Also related: https://github.com/imglib/imglib2/pull/237

src/main/java/net/imglib2/converter/RealTypeConverters.java
maarzt: done

maarzt:
@tpietzsch I did the changes you requested:
* I moved the copy method to `RealTypeConverters.copyFromTo(RandomAccessible source, RandomAccesibleInterval destination)`.
* First argument of the copy method is now `RandomAccessible`.
* I additionally moved all the private methods in `RealTypeConverters` to `RealTypeConverterInternals`. This is to better separate the public methods, from the many private methods. It also provides more space in the `RealTypeConverters` class, to add more functionality like scaling and bounded converters.

tpietzsch:
@maarzt I merged this accidentaly, now force-pushed to master to undo that.
So the state of this PR doesn't reflect reality unfortunately

tpietzsch:
I'll review and then merge locally...

tpietzsch:
ok, merged now...

PR #250, opened 2019-04-08
Multi-Threaded LoopBuilder

maarzt:
This PR improves the performance of `LoopBuilder` by adding two features:
1. `LoopBuilder` now supports multi-threading. This example copies an image using multiple threads:
```java
LoopBuilder.setImages(imageA, imageB).multiThreaded().forEachPixel(
    (a, b) -> b.set(a)
);
```
2. In certain circumstances `LoopBuilder` uses `Cursor` instead of `RandomAccess` to iterate over an image. This happens when all images are `NativeImg` like `ArrayImg`, `PlanarImg` and `CellImg`.

The `multiThreaded()` method in the example above optionally accepts a `MultiThreadSetting` argument:
```java
MultiThreadSetting mts = MultiThreadSettings.multi();
LoopBuilder.setImages(imageA, imageB).multiThreaded(mts).forEachPixel(
    (a, b) -> b. set(a)
);
```
Using the `MultiThreadSetting` one can specify how `LoopBuilder` does the multi-threading. `MultiThreadedSettings.single()` will make the algorithm run without multi-threading, this is also the default. `MultiThreadSettings.multi()` does multi-threading. It's also possible to specify the `ExecutorService` to be used `MultiThreadSettings.multi(16, Executors.newFixedThreadPool(8))`. `MultiThreadSetting` is a very simple interface, have a [look]( https://github.com/imglib/imglib2/blob/multithreaded-loop-builder/src/main/java/net/imglib2/loops/MultiThreadSetting.java). If you want `LoopBuilder` to use a specific thread pool, an custom implementation of `MultiThreadSetting` is the solution.

I think `MultiThreadSetting` can be very useful for image processing algorithms, so I'm happy to discuss it.


src/main/java/net/imglib2/loops/MultiThreadSettings.java
```diff
@@ -0,0 +1,190 @@
...
+public final class MultiThreadSettings
+{
+
+	/**
+	 * Returns a {@link MultiThreadSetting}, that will execute a
+	 * algorithm in a single thread without multi-threading.
+	 */
+	public static MultiThreadSetting single()
+	{
+		return SINGLE_THREAD_SETTINGS;
```
imagejan: Why the plural `S` here?

maarzt: That's a typo, should be without the `S` ;)

ctrueden:
Is it worth looking at using `ForkJoinPool` instead, as discussed in imglib/imglib2-algorithm#81?

maarzt:
```MultiThreadSetting``` is now package-private, and therefore hidden to the user.
We can now merge this PR, to get the multi-threaded ```LoopBuilder```. And after the discussion
about a unified strategy for multi-threading, we can easily change ```MultiThreadSetting``` and ```LoopBuilder```.

@tpietzsch How do you feel about merging this PR?

File: src/main/java/net/imglib2/loops/IntervalChunks.java [hunk 2]
```diff
@@ -0,0 +1,119 @@
...
+	{
+		long[] chunkDimensions = suggestCellSize( Intervals.dimensionsAsLongArray( interval ), numChunks );
+		return generateGrid( interval, chunkDimensions );
+	}
+
+	/**
+	 * Suggest a cell size that can be used, to divide an interval of the
+	 * given dimensions, into the given number of chunks.
+	 */
+	static long[] suggestCellSize( long[] dimensions, long numberOfCells )
```
tpietzsch: I would suggest changing "Cell" to "Chunk" here in parameter, method, and variable names, to make naming consistent

src/test/java/net/imglib2/loops/LoopPerformanceBenchmark.java
```diff
@@ -124,15 +147,18 @@ public void copy_loopBuilder()
 		LoopBuilder.setImages( Views.interval( in, out ), out ).forEachPixel( ( in, out ) -> out.set( in ) );
 	}
 
+	@Benchmark
+	public void copy_flatIterable() {
```
tpietzsch: There is a bug here. It should be `i = Views.flatIterable( Views.interval( in, out ) ).cursor()`, or `o = Views.flatIterable( allOut ).cursor();`. Probably the latter, if you want to benchmark against best-case performance, where the cursors are actually cursors of the underlying ArrayImgs.

src/main/java/net/imglib2/loops/LoopBuilder.java
```diff
@@ -78,68 +83,121 @@
...
+	 */
 	public void forEachPixel( final T action )
 	{
 		Objects.requireNonNull( action );
-		final List< RandomAccess< ? > > samplers = Stream.of( images ).map( this::initRandomAccess ).collect( Collectors.toList() );
-		final Positionable synced = SyncedPositionables.create( samplers );
-		LoopUtils.createIntervalLoop( synced, dimensions, RunnableFactory.bindActionToSamplers( action, samplers ) ).run();
+		if ( Intervals.numElements( dimensions ) == 0 )
+			return;
+		if ( allAreNativeImages() )
```
tpietzsch: This is not a particular good criterion to decide for/against Cursors.
Views can have very fast Cursors, for example if you do a few transformations on an ArrayImg that concatenate to identity, you will get a ArrayImg Cursor (although your input image is a IntervalView or something like that).
It would be better to look at the Cursor types.

src/main/java/net/imglib2/loops/LoopBuilder.java
```diff
@@ -78,68 +83,121 @@
...
+			runUsingCursors( action );
+		else
+			runUsingRandomAccesses( action );
```
tpietzsch: Did you think about mixing (flat) Cursors and RandomAccesses? Maybe not so difficult as it looks at first?

src/main/java/net/imglib2/loops/IntervalChunks.java
maarzt: done

src/main/java/net/imglib2/loops/LoopBuilder.java
maarzt: I tried, but this is not an easy problem: How do I know if a cursor is fast?

maarzt:
@tpietzsch What do you think. Can we merge this?

acardona:
Hi @maarzt,
Question: does this partition the tasks by pixels, that is, two consecutive pixels are processed by different threads? If so, does performance degrade substantially from a larger partitioning approach, whereby instead of processing each pixel independently, a contiguous chunk of the source image is processed by a single thread, to the whole source image is split into as many partitions as there are threads? I suspect the latter has far better performance, given that memory may not be contiguous in large images.

maarzt:
@acardona 
> Question: does this partition the tasks by pixels, that is, two consecutive pixels are processed by different threads?

No, consecutive pixels are usually processed by the same thread. The PR also contains a benchmark that shows multi-threading performance: https://github.com/imglib/imglib2/blob/multithreaded-loop-builder/src/test/java/net/imglib2/loops/LoopPerformanceBenchmark.java
I get a multi-threading speedup of 3.8 on my 4 core machine.

acardona:
Thanks, a 3.8x speed up is wonderful.

Issue #239, opened 2019-01-28
Should Cell implement Interval (or extend AbstractInterval)?


bogovicj:
I found myself wanting to use a `Cell` in `Views.interval( ... )`
Given [this](https://github.com/imglib/imglib2/blob/fb42809c5fa400f1f3a11f43e3931acfb93148ae/src/main/java/net/imglib2/img/cell/Cell.java#L57-L59) I was surprised that it doesn't already. 

Though I presume `Cell`s are supposed to be "small", and I guess this is why their dimensions are stored as an `int[]`?  Because of that, it may not be a very easy change.

Alternatively, could we have a method somewhere that returns an `Interval` from a `Cell`?

tpietzsch:
Good idea. If they implement `Interval` that doesn't add anything to the memory footprint. I wouldn't extend `AbstractInterval`.
A PR would be very welcome. but maybe we wait until #232 is merged...

tpietzsch:
> A PR would be very welcome. but maybe we wait until #232 is merged...
it's merged now...

hanslovsky:
This is probably fixed with #251 and can be closed, no?

PR #251, opened 2019-04-16
Cell now implements Interval

bogovicj:
Only change of note (I think) is that the `dimension` method that formerly returned an `int`, now returns a `long`

see #239


tpietzsch:
thanks @bogovicj

hanslovsky:
I just saw that this is already merged, but this is an interface-breaking change, no? So this would mean a bump of the major version.

tpietzsch:
Hmm, yes... Next time ;-)

hanslovsky:
Fine with me. Sorry I wrote this only after this PR had already been merged.

bogovicj:
and I could have been clearer, oops

PR #254, opened 2019-05-23
Add Method for Unchecked Casts

maarzt:
I suggest to add a method for unchecked casts, that allows to beautifully write:
```java
Img< T > image = Cast.unchecked( ArrayImgs.ints( 100, 100 ) );
```
instead of the more cumbersome:
```java
@SurpressWarnings( "unchecked" )
Img< T > image = ( Img<T> ) ArrayImgs.ints( 100, 100 );
```

### DIFF
--- a/src/main/java/net/imglib2/util/Cast.java
+++ b/src/main/java/net/imglib2/util/Cast.java
@@ -1,0 +1,40 @@
+package net.imglib2.util;
+
+/**
+ * Utility class for type casts.
+ *
+ * @author Matthais Arzt
+ */
+public final class Cast
+{
+
+	private Cast()
+	{
+		// prevent from instantiation
+	}
+
+	/**
+	 * Performs an unchecked cast.
+	 * <p>
+	 * For example this code snipped:
+	 * <pre>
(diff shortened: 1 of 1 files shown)


ctrueden:
A method like this would fit nicely into a SciJava utility package. I am about to factor out some very basic SciJava utility code to a new project `scijava-g3-util` (g3 = "generation 3") with package prefix `org.scijava.g3.util`. This allows forward momentum without breaking backwards compatibility of existing code. The `scijava-g3-util` project will have no dependencies. Would it make sense to put this `Cast` class there instead of in `imglib2`?

tpietzsch:
Would make sense. I also really have no problem with code duplication here. That only becomes problematic if there is two instances of the same interface for example and you would have to decide which to implement/consume...

tpietzsch:
is scijava-g3-util on github yet?

ctrueden:
> is scijava-g3-util on github yet?

I have a local draft of it, but development is stalled. I have too many other things going on, as usual.

Just forget I said anything. It's more pragmatic to work without the status quo for the foreseeable future, rather than planning big restructuring of the codebase.

PR #256, opened 2019-06-18
Generalize acceptable types for image functions

ctrueden:
Since the function populates values by side effect, a `BiConsumer` that works with some more general type than `T` is OK. For example, a function that works on `RealType<?>` could be defined, and then used to create a `FunctionRandomAccessible` with a type creator that is restricted to producing `UnsignedByteType`.

See also [this discussion on Gitter](https://gitter.im/imglib/imglib2?at=5d091817bb789747fb350da9).


axtimwalde:
Thanks!

PR #258, opened 2019-07-09
Deprecate misleading and broken ArrayImgs.bits

ctrueden:
It does not create a `BitType` image. And the `Fraction` part is wrong; the old code may not even work (I didn't test).


ctrueden:
@tpietzsch @axtimwalde @StephanPreibisch Is it OK to self-merge things like this? It is a small change, which I am pretty confident is correct.

tpietzsch:
> Is it OK to self-merge things like this?
From my point of view: Yes.

PR #259, opened 2019-07-19
Checking the validity of input ListIterators

emopers:
Partition.java calls of `j.previous()` on `java.util.ListIterator j` and `i.next()` on `java.util.ListIterator i` without checking if there are any elements to iterate over. Because the method is public and the iterators are obtained from inputs, they could be invalid (e.g., an empty list). This could lead to an exception. This pull request adds a 'hasNext()' and 'hasPrevious()' check. * This PR is a sequel of PR #236 *

### DIFF
--- a/src/main/java/net/imglib2/util/Partition.java
+++ b/src/main/java/net/imglib2/util/Partition.java
@@ -37,6 +37,7 @@ package net.imglib2.util;
 import java.util.Comparator;
 import java.util.List;
 import java.util.ListIterator;
+import java.util.NoSuchElementException;
 
 /**
  * TODO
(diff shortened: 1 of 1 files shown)


ctrueden:
Thanks @emopers!

However, after seeing the same sorts of changes in #236 and #260, I am beginning to think this sort of change might become annoying to people reading the tests. Adding these tests to every place where cursors are used bogs down the logic of the tests. Properly, we should be testing the cursor iteration in its own test method, and then in other tests, we can assume the iteration works properly.

In other words: it is not the job of every test to test every piece of logic used in the execution of that test.

@emopers Did you find all these places using some code linting tool? I'm not sure I agree that the lack of `hasNext` checks in these circumstances are "violations"...

@tpietzsch What do you think? Should we merge #236, #259 and #260? Or close them without merge?

tpietzsch:
@ctrueden I would close #259 and #260 without merge.
I agree with the intention of #236, but it should rather check `i.nextIndex() < j.previousIndex()` or something (I would need to think about the exact condition). Unfortunately, there is no way to check whether `i` and `j` are iterators into the same list, which is also an expectation about the input.

PR #260, opened 2019-07-19
Checking the validity of input Cursors

emopers:
These tests calls of `Cursor.next()` without checking if there are any elements to iterate over. Because the method is public and the iterators are obtained from inputs, they could be invalid (e.g., an empty list). This could lead to an exception. This pull request adds a 'hasNext()' check.

### DIFF
--- a/src/test/java/net/imglib2/img/cell/CopyTest.java
+++ b/src/test/java/net/imglib2/img/cell/CopyTest.java
@@ -133,7 +133,7 @@ public class CopyTest
 		{
 			if ( !dst.hasNext() )
 			{
-				throw new NoSuchElementException("Cursor does not have next element");
+				throw new NoSuchElementException(" Cursor dst does not have next element" );
 			}
 			dst.next().set( src.next().get() );
 		}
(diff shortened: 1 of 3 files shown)


src/test/java/net/imglib2/img/cell/CopyTest.java
```diff
@@ -130,6 +131,10 @@ public void copyWithIterationBoth( final Img< IntType > srcImg, final Img< IntTy
 		final Cursor< IntType > dst = dstImg.cursor();
 		while ( src.hasNext() )
 		{
+			if ( !dst.hasNext() )
+			{
+				throw new NoSuchElementException("Cursor does not have next element");
```
ctrueden: Would it not be better to call JUnit's `fail` method here, for consistency? These methods, while public, are only intended for use from unit tests, right?

src/test/java/net/imglib2/view/RandomAccessibleIntervalCursorTest.java
```diff
@@ -103,7 +104,14 @@ public void setUp()
 	public void copy( final Cursor< IntType > src, final Cursor< IntType > dst )
 	{
 		while ( src.hasNext() )
+		{
+			if ( !dst.hasNext() )
+			{
+				throw new NoSuchElementException("Cursor dst does not have next element");
```
ctrueden: Similarly: should we use `fail` here?

ctrueden:
Upon further reflection, I am ambivalent about this. See https://github.com/imglib/imglib2/pull/259#issuecomment-521686798

PR #266, opened 2019-08-26
Add toString, equals and hashCode for Intervals, RealIntervals & Points

maarzt:
It would be super useful for debugging and development, if we have suitable `toString`, `equals` and `hashCode` methods for `Intervals` and `Points`. This PR add methods:

* `toString()`, `hashCode()` and `equals(Object)` for `FinalInterval`, `FinalRealInterval`, `Point`, `RealPoint` and `FinalDimensions`, if missing.
* `Intervals.equals(RealInterval, RealInterval)`
* `Intervals.equals(RealInterval, RealInterval, double tolerance)`
* `Intervals.equalsDimensions(Dimensions, Dimensions)`
* `Intervals.toString(Interval)`
* `Intervals.toString(RealInterval)`
* `Intervals.toString(Dimensions)`
* `Localizables.equals(Localizable, Localizable)`
* `Localizables.equals(RealLocalizable, RealLocalizable)`
* `Localizables.equals(RealLocalizable, RealLocalizable, double tolerance)`
* `Localizables.toString(Localizable)`
* `Localizables.toString(RealLocalizable)`

Sorry, this PR is based on a previous PR https://github.com/imglib/imglib2/pull/265

### DIFF
--- a/src/main/java/net/imglib2/AbstractInterval.java
+++ b/src/main/java/net/imglib2/AbstractInterval.java
@@ -34,6 +34,8 @@
 
 package net.imglib2;
 
+import net.imglib2.util.Intervals;
+
 /**
  * Implementation of the {@link Interval} interface.
  * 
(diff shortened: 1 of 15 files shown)


tpietzsch:
> Changing `Point.equals()` to
> 
> ```java
> @Override
> public boolean equals( final Object obj )
> {
> 	return obj != null && obj.getClass().equals( getClass() ) && Localizables.equals( this, ( Localizable ) obj );
> }
> ```
> 
> fixes that.

If nobody has objections, I'll just fix it like that and merge?

ctrueden:
Sounds good to me. Note that Class does not override equals(Object), so it
would be sufficient to check `obj.getClass() == getClass()`.

>

maarzt:
Sounds good to me too.

Issue #210, opened 2018-05-21
Views.expandValue and Converters.convert bounds too strict


hanslovsky:
Some methods and classes take `Type< T >` as parameter when instead they should take `Supplier< T >`. This results in an unnecessary restriction of the generic parameter to `Type< T >` of `Views.extendValue` and `Converters.convert`. There may be other classes and methods but these are the ones that I noticed. My specific use case is extending/converting `RandomAccessibleInterval< Volatile< BooleanType > >`, similar to this MWE:
```java
public static void main( final String[] args )
{
    final BoolType t = new BoolType( true );
    final BoolType f = new BoolType( false );
    final Volatile< BoolType > vt = new Volatile<>( t, true );
    final Volatile< BoolType > vf = new Volatile<>( f, true );
    final RandomAccessibleInterval< Volatile< BoolType > > source = ConstantUtils.constantRandomAccessibleInterval( vt, 2, new FinalInterval( 10, 20 ) );
    
    // try to extendValue
    Views.extendValue( source, vf );
    
    // try to convert
    final Converter< Volatile< BoolType >, Volatile< BoolType > > conv = ( src, tgt ) -> tgt.get().set( !src.get().get() );
    Converters.convert( source, conv, vf );
}
```
Compiler errors:
```
The method extendValue(F, T) in the type Views is not applicable for the arguments (RandomAccessibleInterval<Volatile<BoolType>>, Volatile<BoolType>)

The method convert(RandomAccessible<A>, Converter<? super A,? super B>, B) in the type Converters is not applicable for the arguments (RandomAccessibleInterval<Volatile<BoolType>>, Converter<Volatile<BoolType>,Volatile<BoolType>>, Volatile<BoolType>)
```
 
Of course, in this example, 
 - `Volatile< BooleanType >` could be replaced by `VolatileUnsignedByteType` and a non-zero check on the value of `VolatileUnsignedByteType`. However, for a binary mask, having an actual boolean value (instead of a numeric value) is more descriptive of what is happening and leads to more understandable code. Pseudo code:
 ```java
 UnsignedByteType u = //
 u.set( someCondition ? 1 : 0 );
 isTrue = u.get() == 1;
 
 BooleanType b = //
 b.set( someCondition );
 isTrue = b.get();
 ```
 Or
  - implement a type that extends `Volatile` and implements `BooleanType`. Unfortunately, that is not possible at the moment(#209).
  
[...]

PR #268, opened 2019-09-04
Remove generic bounds for constant out of bounds

hanslovsky:
Addresses #210 (at least reagarding `Views.extend/expand`; if these changes are OK, I will adapt the respective methods in `Views` as well before this PR is merged).
Replaces #194

As far as I can tell, this does not break any interface wrt the constructors and the `setValue` method.

However, this technically breaks public interface of `OutOfBoundsConstantValueFactory`:
 1. The protected member `OutOfBoundsConstantValueFactory.value` is replaced with `OutOfBoundsConstantValueFactory.valueSupplier`.
 2. `OutOfBoundsConstantValueFactory.getValue().set(...)` does not have any effect on the internal state (not an actual interface break)

TODO:
 - [x] Use raw `Type` for `t instanceof Type` [check](https://github.com/imglib/imglib2/pull/268#discussion_r321126026).
 - [x] Update `Views.extendValue` and `Views.expandValue`.

### DIFF
--- a/src/main/java/net/imglib2/outofbounds/OutOfBoundsConstantValue.java
+++ b/src/main/java/net/imglib2/outofbounds/OutOfBoundsConstantValue.java
@@ -98,8 +98,10 @@ public class OutOfBoundsConstantValue< T > extends AbstractOutOfBoundsValue< T >
 		return copy();
 	}
 
-	private static < T > Supplier< T > makeSupplierFrom( final T t ) {
-		if ( t instanceof Type< ? > ) {
+	private static < T > Supplier< T > makeSupplierFrom( final T t )
+	{
+		if ( t instanceof Type< ? > )
+		{
 			final Type< ? > type = ( Type< ? > ) t;
 			return () -> ( T ) type.copy();
 		}
--- a/src/main/java/net/imglib2/outofbounds/OutOfBoundsConstantValue.java
+++ b/src/main/java/net/imglib2/outofbounds/OutOfBoundsConstantValue.java
@@ -88,12 +88,14 @@ public class OutOfBoundsConstantValue< T > extends AbstractOutOfBoundsValue< T >
 		return copy();
 	}
 
-	private static < T > T copyIfType( final T t )
+	static < T > T copyIfType( final T t )
 	{
 		if ( t instanceof Type< ? > )
 		{
 			final Type< ? > type = ( Type< ? > ) t;
-			return ( T ) type.copy();
+			@SuppressWarnings( "unchecked" )
+			final T copy = ( T ) type.copy();
+			return copy;
 		}
 		return t;
 	}
--- a/src/main/java/net/imglib2/outofbounds/OutOfBoundsConstantValue.java
+++ b/src/main/java/net/imglib2/outofbounds/OutOfBoundsConstantValue.java
@@ -60,7 +60,7 @@ public class OutOfBoundsConstantValue< T > extends AbstractOutOfBoundsValue< T >
 	public < F extends Interval & RandomAccessible< T > > OutOfBoundsConstantValue( final F f, final T value )
 	{
 		super( f );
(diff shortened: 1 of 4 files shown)


src/main/java/net/imglib2/outofbounds/OutOfBoundsConstantValue.java
```diff
@@ -85,4 +97,12 @@ final public T get()
...
 		return copy();
 	}
+
+	private static < T > Supplier< T > makeSupplierFrom( final T t ) {
+		if ( t instanceof Type< ? > ) {
+			final Type< ? > type = ( Type< ? > ) t;
+			return () -> ( T ) type.copy();
+		}
+		return () -> t;
+	}
```
tpietzsch: Please use imglib2 formatter

src/main/java/net/imglib2/outofbounds/OutOfBoundsConstantValue.java
```diff
@@ -85,4 +97,12 @@ final public T get()
 	{
 		return copy();
 	}
+
+	private static < T > Supplier< T > makeSupplierFrom( final T t ) {
+		if ( t instanceof Type< ? > ) {
```
tpietzsch: I would make this `t instanceof Type` (instead of `Type< ? >`)
It's kind of an edge case, but for example `t instanceof Type< ConcreteType >` doesn't compile, because instanceof doesn't check generic parameters.

tpietzsch:
I don't think the protected `value` is used anywhere. Maybe we make `value` and `valuesupplier` private while we (technically) break API. In any case, I wouldn't bump major version because of this...

src/main/java/net/imglib2/outofbounds/OutOfBoundsConstantValue.java
hanslovsky: Sorry, my bad. Fixed in 13b05322174be24818497924f42b189f163dff14

src/main/java/net/imglib2/outofbounds/OutOfBoundsConstantValue.java
hanslovsky: Should I update the cast to raw type as well (`( Type )` instead of `( Type< ? > )`?

hanslovsky:
> Maybe we make value and valuesupplier private while we (technically) break API.

Definitely in favor of that.

> In any case, I wouldn't bump major version because of this...

agreed

src/main/java/net/imglib2/outofbounds/OutOfBoundsConstantValue.java
ctrueden: I think the cast should still be to `Type< ? >` because that's the properly specified generic type here. You don't know anything about the bound; writing `Type< ? >` is the most general expression for that without introducing type variables.

hanslovsky:
I think this can be greatly simplified without the `Supplier` and by using a function like `copyIfType` instead:
```java
private static < T > T copyIfType( final T t )
{
	if ( t instanceof Type< ? > )
	{
		final Type< ? > type = ( Type< ? > ) t;
		return ( T ) type.copy();
	}
	return t;
}
```
The `copyIfType` method would be less performant than the supplier (because it has to type check and dynamically cast everytime) but it would be called only at construction or `setValue`, so only once, really. The code, however, would be much simpler and pretty much the same as before. I will push the changes.

hanslovsky:
The changes now really boil down to only:
 - add `copyIfType` method
 - replace `Type.copy` with `copyIfType`
 - relax generic bounds.

tpietzsch:
@hanslovsky There is now two identical static `copyIfType` methods (in `OutOfBoundsConstantValueFactory` and `OutOfBoundsConstantValue`).
Please remove one of them and reuse the other (make it package private).

Then this PR can be squashed and merged.

Issue #240, opened 2019-01-28
Convenient methods for iterating over cells / blocks


bogovicj:
@axtimwalde @hanslovsky @igorpisarev @tpietzsch  

This got kind of long, sorry, but mostly I just want to start a discussion.

There are a number of situations in which one may not have a `CellImg` explicitly but want to iterate over blocks.  The answer today (afaik) seems to be: "keep track of everything on your own."  Unless there exists something already that I don't know about (@hanslovsky ?)

Example use case, we often have to explicitly [enumerate/index the blocks](https://github.com/saalfeldlab/n5-spark/blob/456f1d9029bb886142185d96e6fe1cb0a74a8276/src/main/java/org/janelia/saalfeldlab/n5/spark/N5ConvertSpark.java#L241-L242), and later [build the intervals](https://github.com/saalfeldlab/n5-spark/blob/456f1d9029bb886142185d96e6fe1cb0a74a8276/src/main/java/org/janelia/saalfeldlab/n5/spark/N5ConvertSpark.java#L246-L255) that represent the relevant block.  This pattern gets [repeated](https://github.com/saalfeldlab/n5-spark/blob/master/src/main/java/org/janelia/saalfeldlab/n5/spark/downsample/N5DownsamplerSpark.java#L124-L144) a bunch.

In my opinion, it would be very nice to be able to get a `Iterable<Interval>` or `Cursor<Interval>` [1][2]. then: 
```
CellGrid grid = ...
RandomAcessibleInterval someimg  = ... // not necessarily a CellImg

Cursor<Interval> = grid.cursor();
while( grid.hasNext())
{
   block = Views.interval( someimg, grid.next());
   // do block processing 
}
```
which is quite a bit more concise than the current situation (as I understand it).

[1] or maybe `Cursor<Cell>`? (see also #239)
[2] makes sense to me for `CellGrid`s to provide it

axtimwalde:
Here are my 5ct to this.  As you can see, it's messy because each of these methods lump a lot of information into a single 'cell' and try to avoid dedicated classes which is bad design.  But it's a good overview of stuff that I needed in this particular project, i.e. may be a good reminder for what to consider (e.g. that the block-size is not necessarily the grid spacing, padding, trailing cells etc.).

hanslovsky:
[...]

PR #269, opened 2019-09-19
Parallelization Framework for ImgLib2

maarzt:
This is the realization of an idea how to organize imglib2 related multi-threaded code. I previously discussed it with @tpietzsch, @hanslovsky, @axtimwalde. Your thoughts are very welcome.

The two concepts in this PR, are the interface `TaskExecutor` and the class `Parallelization`:

**TaskExecutor**

`TaskExecutor` actually is an interface not a class. It's similar to `ExecutorService` but offers simpler methods. It's better suited to image processing algorithms. The following example uses `TaskExecutor` to fill an image with all ones in a multi-threaded way:
```java
public void fillImageWithOnes( final RandomAccessibleInterval< IntType > image,
                               final TaskExecutor taskExecutor )
{
    int numTasks = taskExecutor.suggestNumberOfTasks();
    List< RandomAccessibleInterval< IntType > > chunks =
        splitImageIntoChunks( image, numTasks );

    // The TaskExecutor executes the forEach method in multi threads, if requested.
    taskExecutor.forEach( chunks, chunk -> {
        for ( IntType pixel : Views.iterable( chunk ) )
            pixel.setOne();
    } );
}
```
This method can be called by `fillImageWithOnes( image, TaskExecutors.multiThreaded() )`. And it's possible to run it single threaded by `fillImageWithOnes( image, TaskExecutors.singleThreaded() )`. 

`TaskExecutor.singleThreaded()` is very lightweight. It requires now resources, threads, or memory allocation, and can therefor also be used in very tight loops. When single threaded `taskExecutor.suggestNumberOfTasks()` returns 1, and `taskExecutor.forEach(...)` is just a simple for loop. This means that `splitImageIntoChunks( image, 1 )` can just return the original image, and as a consequence `fillImageWithOnes` will run with optimal performance, in both modes: single- and multi-threaded.

Lets imagine, what the method would look like, if `ExecutorService` is used:
1. It would need to guess how many chunks are usefull, maybe take: 
`Runtime.getRuntime().availableProcessors() * 4`.
2. A list of chunks need to be created.
3. For each chunk we need to generate a `Callable<Void>`.
4. Call `Future<Void> futeres = executorService.invokeAll(callables)`.
5. Iterate over all futures, and call `future.get()'.
6. Handle `ExecutionException` and `ItenterruptedException'.

The `ExecutorService` based function is way more complicated. To run it single threaded one would use `Executors.newSingleThreadedExecutor()`, which has a big overhead.

**Parallelization**

In ImgLib2 it's common that a multi-threaded algorithm has some additional parameters, `ExecutorService`, `numThreads` or `numberOfTask` or combinations of these. This means the user needs to specify an `ExecutorService`, or number of threads. But most time a user wouldn't want to care about that, and just run the method quickly.

I suggest the following approach:
```java

// If you don't care, just run:
fillImageWithOnes( image );

// Single Threaded call
Parallelization.runSingleThreaded( () -> fillImageWithOnes( image ) );

// Multi Threaded call
Parallelization.runMultiThreaded( () -> fillImageWithOnes( image ) );

// Call with a specific ExecutorService
Parallelization.runWithExecutor( executorService, () -> fillImageWithOnes( image ) );

// Call with the sepecified number of threads
Parallelization.withNumThreads( numThreads, () -> fillImageWithOnes( image ) );

// Example Algorithm, that fills an image with ones.
public void fillImageWithOnes( RandomAccessibleInterval< IntType > image )
{
    TaskExecutor taskExecutor = Parallelization.getTaskExecutor();
    int numTasks = taskExecutor.suggestNumberOfTasks();
    List< RandomAccessibleInterval< IntType > > chunks = splitImageIntoChunks( image, numTasks );

    // The TaskExecutor executes the forEach method in multi threads, if requested.
    taskExecutor.forEach( chunks, chunk -> {
        for ( IntType pixel : Views.iterable( chunk ) )
            pixel.setOne();
    } );
}
```

The class `Parallelization` associates a `TaskExecutor` with each thread. (`ThreadLocal<TaskExecutor>` is used internally.) `Parallelization.getTaskExecutor()` returns the task executor for the current task. `Parallelization.runSingleThreaded( task )` first sets the task executor of the current thread to `TaskExecutor.singleThreaded()`, then executes the task, and finally restores the current threads task executor to it's original value. The other methods `runMultiThreaded` and `runWithExecutor` work similarly.

This approach has several advantage:
1. The `TaskExecutor` doesn't need to be passed around as a parameter.
2. There is a reasonable default value. Which will use the common `ForkJoinPool`.
3. It makes sense to associate, a `TaskExecutor` setting with a thread. The threads in a fixed thread pool ('Executors.newFixedThreadPool()`) shouldn't do nested parallelization. That's way these threads need their task executors be set to `TaskExecutor.singleThreaded()`. (This is done for you just use `TaskExecutors.fixedThreadPool()`).

**LoopBuilder & ImgLib2-Algorithm**

I used this approach in `LoopBuilder`. Let's write the `fillImageWithOnes` method once more:
```java
public void fillImageWithOnes( RandomAccessibleInterval< IntType > image )
{
    LoopBuilder.setImages( image ).multithreaded().forEachPixel( pixel -> pixel.setOne() );
}
```
It's possible to execute this method single, and multi-threaded with the `Parallelization.runSingleThreaded()`, etc.

I used the presented approach, in imglib2-algorithm. I will put a link below.

**Backwards compatibility**

There are a whole bunch of methods, that ensure compatibility with the other approaches:
```java
executorService = taskExecutor.getExecutorService(); // TaskExecutor -> ExecutorService
taskExecutor = TaskExecutors.forExecutorService(executorService); // ExecutorService -> TaskExecutor
numThreads = taskExecutr.getParallelism(); // TaskExector -> numThreads
taskExecutor = TaskExecutors.numThreads(numThreads); // numTreads -> taskExecutor
taskExecutor = TaskExecutors.forExecutorServiceAndNumTasks(executorService, numTasks); // ExecutorService & numTasks -> TaskExecutor
```


src/main/java/net/imglib2/parallel/DefaultTaskExecutor.java
```diff
@@ -0,0 +1,183 @@
...
+		for ( Runnable task : tasks )
+			callables.add( Executors.callable( task ) );
+		invokeAllIgnoreResults( callables );
+	}
+
+	@Override
+	public int suggestNumberOfTasks()
+	{
+		int parallelism = getParallelism();
+		return ( parallelism == 1 ) ? 1 : ( int ) Math.min( ( long ) parallelism * 4L, ( long ) Integer.MAX_VALUE );
```
hanslovsky: The max size of a Java array (or an `ArrayList`) is a little smaller than `Integer.MAX_VALUE` and the exact size depends on the JVM. For my JVM (`openjdk version "1.8.0_222"`), for example,
```kotlin
val list = ArrayList<Unit?>(Integer.MAX_VALUE)
```
fails with
```
Caused by: java.lang.OutOfMemoryError: Requested array size exceeds VM limit
        at java.util.ArrayList.<init>(ArrayList.java:153)
        at Max_int_array_list.<init>(Unknown Source)
        ... 13 mor
```

The max size that does not fail in my JVM is `Integer.MAX_VALUE - 3`:
```kotlin
val list = ArrayList<Unit?>(Integer.MAX_VALUE - 3)
```

I am not aware of a way to retrieve the VM limit but nonetheless it would be good to limit the size to a little smaller than `Integer.MAX_VALUE`, e.g. `Integer.MAX_VALUE - 5`.

https://bugs.openjdk.java.net/browse/JDK-8059914?focusedCommentId=13585092&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13585092

src/main/java/net/imglib2/parallel/Parallelization.java
```diff
@@ -0,0 +1,266 @@
...
+ *
+ *     // ExecutorService
+ *     Parallelization.withExecutor( executorService ).run( () -> myAlgorithm( image ) );
+ *
+ *     // Example Algorithm, that fills an image with ones.
+ *     public void myAlgorithm( RandomAccessibleInterval< IntType > image )
+ *     {
+ *         TaskExecutor taskExecutor = Parallelization.getTaskExecutor();
+ *         int numTasks = taskExecutor.suggestNumberOfTasks();
+ *         List< RandomAccessibleInterval< IntType > > chunks = splitImageIntoChunks( image, numTasks );
```
hanslovsky: Please add a note in here that it is the user's responsibility to create a method `splitImageIntoChunks`, otherwise people might be looking for that method in the imglib2 code base. Maybe also add a comment that chunking may be task specific, e.g. for separable distance transform.

hanslovsky:
@maarzt this looks good, great work! I added a few comments to the code. Additionally, I think that a method
```java
Optional<int[]> TaskExecutor.suggestChunkSize(Dimensions dimensions)
```
would be super useful if a certain block/chunk size is required, e.g. when writing to N5 in `myAlgorithm`. We should probably discuss here, if chunk size should be a parameter of `myAlgorithm` or a method in the `TaskExecutor` interface.

src/main/java/net/imglib2/parallel/DefaultTaskExecutor.java
ctrueden: This is also a problem with `Util.getArrayOrCellImgFactory` currently, no? Because we use `Integer.MAX_VALUE` for the cutoff between `ArrayImg` and `CellImg`.

FWIW, on my macOS system, the maximum array size seems to be `Integer.MAX_VALUE - 2` consistently. Tested empirically with the following AdoptOpenJDK versions:  12, 11.0.2, 10.0.2, 9, 1.8.0_212.

ctrueden:
@maarzt Thanks a lot for working on this!
@gselzer You may be interested in this as well.

src/main/java/net/imglib2/parallel/Parallelization.java
maarzt: Lets add the `splitImageIntoChunks` method to imglib2. It would be great to have it!?

maarzt:
@hanslovsky Thank you for reviewing this so quickly! :heart:

Please elaborate a little more on this `suggestChunkSize(...)` method. How would it work? 
Could this be a static method in another class `Chunker.suggestChunkSize( Dimensions size, int numTasks )`?

@gselzer Do I remember correctly, that you are working on Ops? I am really curios, if this PR works well together with Ops. Could it be used for parallelization there?

src/main/java/net/imglib2/parallel/Parallelization.java
acardona: I second that--having a utility method view an image as chunks would facilitate lots of things beyond parallel processing (but particularly parallel processing). Will need an out of bounds strategy that accesses the underlying ```RandomAccessibleInterval```, though.

src/main/java/net/imglib2/parallel/DefaultTaskExecutor.java
hanslovsky: Yes, that would be a problem there, too.

src/main/java/net/imglib2/parallel/Parallelization.java
axtimwalde: > I second that--having a utility method view an image as chunks would facilitate lots of things beyond parallel processing (but particularly parallel processing). Will need an out of bounds strategy that accesses the underlying `RandomAccessibleInterval`, though.

In my world, chunks almost always need context but I would almost never out-of-bounds-pad them if there is real context available.  I am therefore hesitant to think this method is generally useful for anything but per-pixel processing?  I always work with a set of plain Intervals generated by a customizable chunking method and passing along the original (out-of-bounds-padded) source and the Interval, so that the consuming method can do what is right with it (e.g. make an IntervalView).

src/main/java/net/imglib2/parallel/DefaultTaskExecutor.java
tpietzsch: Yes, we should change that. Actually I also had that on my TODO list already...

In Mastodon, we use [this:](https://github.com/bigdataviewer/mastodon-graph/blob/f432bf7e8a5784e1e8ed079672a060599e78fe63/src/main/java/org/mastodon/undo/UndoDataStack.java#L113-L118)
```java
	/**
	 * The maximum size of array to allocate. Some VMs reserve some header
	 * words in an array. Attempts to allocate larger arrays may result in
	 * OutOfMemoryError: Requested array size exceeds VM limit
	 */
	private static final int MAX_ARRAY_SIZE = Integer.MAX_VALUE - 8;
```
We could put it into a `net.imglib2.Constants` class and then use everywhere where we use `Integer.MAX_VALUE` at the moment. But as @ctrueden suggested, maybe we should make it a method instead of a field, so that it will not be inlined as a compile time constant.

src/main/java/net/imglib2/parallel/Parallelization.java
tpietzsch: > I am therefore hesitant to think this method is generally useful for anything but per-pixel processing?

I tend to agree. But I'm open to suggestions!
Maybe a full-fledged generally useful `splitImageIntoChunks` is not possible, but certainly little helpers could be added to make special case implementations easier.

These discussions are related:
https://github.com/imglib/imglib2/issues/240
https://github.com/imglib/imglib2/pull/179
and maybe more...?

src/main/java/net/imglib2/parallel/DefaultTaskExecutor.java
ctrueden: A constant might be OK as long as it is `private`. Because the only things that can use it then will be in the same component, and will also be recompiled if it ever changes.

src/main/java/net/imglib2/loops/LoopBuilder.java
```diff
@@ -183,7 +224,18 @@ private boolean cursorIsFast( IterableInterval<?> image )
 	 */
 	public LoopBuilder< T > multiThreaded()
 	{
-		this.multiThreaded = Objects.requireNonNull( MultiThreadSetting.MULTI );
+		return multiThreaded( Parallelization.getTaskExecutor() );
```
tpietzsch: Please add clarification to the javadoc that this will use multi-threading if run in a parallelized context.
(If I understand correctly, if run using `Parallelization.runSingleThreaded( ... )` then `multiThreaded()` does nothing?)

src/main/java/net/imglib2/parallel/DefaultTaskExecutor.java
```diff
@@ -0,0 +1,183 @@
...
+	{
+		if ( executorService instanceof ForkJoinPool )
+			return ( ( ForkJoinPool ) executorService ).getParallelism();
+		else if ( executorService instanceof ThreadPoolExecutor )
+			return Math.max( 1, ( ( ThreadPoolExecutor ) executorService ).getCorePoolSize() );
+		else if ( executorService instanceof ForkJoinExecutorService )
+			return ( ( ForkJoinExecutorService ) executorService ).getParallelism();
+		else if ( executorService instanceof SequentialExecutorService )
+			return ( ( SequentialExecutorService ) executorService ).getParallelism();
+		return Runtime.getRuntime().availableProcessors();
```
tpietzsch: this `instanceof` cascade could be done once on construction to create a reusable `IntSupplier`

src/main/java/net/imglib2/parallel/DefaultTaskExecutor.java
```diff
@@ -0,0 +1,183 @@
...
+	{
+		int parallelism = getParallelism();
+		return ( parallelism == 1 ) ? 1 : ( int ) Math.min( ( long ) parallelism * 4L, ( long ) Integer.MAX_VALUE );
+	}
+
+	@Override
+	public < T > void forEach( List< ? extends T > parameters, Consumer< ? super T > task )
+	{
+		List< Callable< Object > > callables = new ArrayList<>( parameters.size() );
+		// use for-loop because stream with collect(Collectors.toList) is slow.
```
tpietzsch: ❤️

src/main/java/net/imglib2/parallel/ForkJoinExecutorService.java
```diff
@@ -0,0 +1,146 @@
...
+	@Override
+	public < T > Future< T > submit( Runnable runnable, T t )
+	{
+		return ForkJoinTask.adapt( runnable, t ).fork();
+	}
+
+	@Override
+	public < T > Future< T > submit( Callable< T > callable )
+	{
+		return ForkJoinTask.adapt( callable ).fork();
```
tpietzsch: @maarzt Can you elaborate a bit on `ForkJoinExecutorService` design?

`ForkJoinPool` implements `ExecutorService`.
So this could also be implemented as
`return getPool().submit( callable )`.
And similar for the other methods.
Is there a reason to do it like this? What are the trade-offs?
I'm not getting the full picture here...

src/main/java/net/imglib2/parallel/Parallelization.java
tpietzsch: Also related is `net.imglib.loop.IntervalChunks`, but it's not very flexible

src/main/java/net/imglib2/parallel/Parallelization.java
```diff
@@ -0,0 +1,266 @@
...
+	 *     myAlgorithm(input);
+	 * }
+	 * finally
+	 * {
+	 *     frame.close();
+	 * }
+	 * }
+	 * </pre>
+	 */
+	public static Frame setExecutorRequiresReset( TaskExecutor taskExecutor )
```
tpietzsch: I'm not sure this is needed/should be public.
The example in the javadoc could be just `runWithExecutor(taskExecutor, () -> myAlrogithm(input));`?
Where is this useful?

src/main/java/net/imglib2/parallel/Parallelization.java
```diff
@@ -0,0 +1,266 @@
...
+	{
+		final TaskExecutor old = executor.get();
+		executor.set( taskExecutor );
+		return () -> executor.set( old );
+	}
+
+	public interface Frame extends AutoCloseable
```
tpietzsch: It's confusing that Frame has a `void close()` and `void reset()` method that do the same thing.
Why not simply use `close()`? (You even suggest to use close() in the `setExecutorRequiresReset` javadoc)
I.e., this could be simply `public interface Frame extends AutoCloseable {}`.

src/test/java/net/imglib2/parallel/DefaultTaskExecutorTest.java
```diff
@@ -0,0 +1,142 @@
...
+	public void testForEach()
+	{
+		AtomicInteger sum = new AtomicInteger();
+		List< Integer > parameters = Arrays.asList( 1, 2, 3 );
+		twoThreads.forEach( parameters, sum::addAndGet );
+		assertEquals( 6, sum.get() );
+	}
+
+	@Test
+	public void testForEachCollectResults()
```
tpietzsch: Rename method to `testForEachApply()`?

tpietzsch:
@maarzt very nice! Besides minor issues/questions (see inline comments) I like `net.imglib2.parallel` a lot. Because it is completely independent of other imglib2 stuff, we could consider moving it into a common `scijava-parallel` artefact (which would have no dependencies)? Opinions? (@ctrueden: Do you think this would be useful anywhere upstream of where `imglib2` comes in?)

LoopBuilder changes are also fine.

src/main/java/net/imglib2/parallel/Parallelization.java
hanslovsky: I agree with @axtimwalde and would like to add that there is the [`Grids`](https://github.com/imglib/imglib2-algorithm/blob/47cd6ed5c97cca4b316c92d4d3260086a335544d/src/main/java/net/imglib2/algorithm/util/Grids.java) utility class to generate lists of `Interval` from `Dimensions` and `blockSize`.

hanslovsky:
> Please elaborate a little more on this `suggestChunkSize(...)` method. How would it work?
> Could this be a static method in another class `Chunker.suggestChunkSize( Dimensions size, int numTasks )`?

In many of our tasks, the way we parallelize is ultimately determined by the block/chunk/cell size of the data that we write out, eventually. In particular, that means that we need to provide a `blockSize` to the method and the data is processed in chunks of (multiples of) `blockSize`. This `blockSize` can now be either
 1. be passed as explicit parameter to the method
 2. be provided by the `TaskExecutor` object.

It may be preferable to always pass it as explicit parameter to the method anyway, so callers are aware that they need to set this parameter explicitly, instead of falling back to a default value (if any) that does make sense for the number of tasks, but not for the block/cell/chunk size.

This may also be more of a concern for distributed memory computations.

hanslovsky:
This probably makes the [`ParallelizeOverBlocks`](https://github.com/imglib/imglib2-algorithm/blob/47cd6ed5c97cca4b316c92d4d3260086a335544d/src/main/java/net/imglib2/algorithm/util/ParallelizeOverBlocks.java) class in imglib2-algorihtm obsolete

ctrueden:
> Do you think this would be useful anywhere upstream of where imglib2 comes in?

Yeah, maybe the scijava-ops framework should make use of it. Hopefully I'll have time to do some experiments before the hackathon is over.

src/main/java/net/imglib2/loops/LoopBuilder.java
maarzt: done

src/main/java/net/imglib2/parallel/DefaultTaskExecutor.java
maarzt: Done. I also measure performance, and it reduces execution time by two.

src/main/java/net/imglib2/parallel/Parallelization.java
maarzt: Done. I removed the reset method.

src/test/java/net/imglib2/parallel/DefaultTaskExecutorTest.java
maarzt: done.

src/main/java/net/imglib2/parallel/DefaultTaskExecutor.java
maarzt: Sorry it the opposite. I red the performance measurement results wrongly. Using `IntSupplier` results in less performance. So I removed it again.

src/main/java/net/imglib2/parallel/ForkJoinExecutorService.java
maarzt: We need `ForkJoinExecutorService` because it enables nested parallelization. Using `ForkJoinPool` for nested parallelization at some point throws an RejectedExecutionException: "Thread limit exceeded replacing blocked worker".

This happens, because `forkJoinPool.invokeAll(...)` is called inside a `ForkJoinThread`, which will block this thread until all the tasks are completed. The same is true for a call to `forkJoinPool.submit(...).get()`. The blocked thread can no longer be used for task execution. Which is bad as this wastes resources. `ForkJoinPool` mitigates this by creating new threads, that's what after some time causes the exception "Thread limit exceeded...".

So when you're inside a `ForkJoinPool` it's a bad idea to call `forkJoinPool.invokeAll` or `forkJoinPool.submit`. One should instead use `forkJoinTask.fork()` or `forkJoinTask.invokeAll()`.
And that's exactly what `ForkJoinExecutorService` does: it uses `ForkJoinTask.invokeAll(...)` instead of `forkJoinPool.invokeAll(...)`. And that makes `ForkJoinExecutorService` so much better for nested parallelization. 

And surprisingly `ForkJoinTask.invokeAll(...)` and `forkJoinTask.submit()` work anywhere, even outside a `ForkJoinPool`. That's why it's OK to use the `ForkJoinExecutorService` anywhere.

This advantages, have a downside too, `ForkJoinExecutorService` is not a fully functional `ExecutorService`, methods like `awaitTerminaten` or `shutdownNow` don't work.

src/main/java/net/imglib2/parallel/Parallelization.java
maarzt: This is useful, if wrapping the method call into a lambda is not an option. We could for example write a `@PreprocesserPlugin` and a `@PostprocessorPlugin` for SciJava. The  `@PreprocesserPlugin` plugin would call this method, to set the `TaskExecutor` for a `Command`, and the `@PostprocessorPlugin` would close the frame.

But the are other options too, the `Command` could use `TaskExector.setExecutor(...)`, or maybe the `ModuleService` would do so. @ctrueden What are your thought's on this?

src/main/java/net/imglib2/parallel/Parallelization.java
maarzt: > I am therefore hesitant to think this method is generally useful for anything but per-pixel processing?

That's probably correct. And LoopBuilder could be used for per-pixel processing instead. That makes the `splitImageIntoChunks` method rather useless. I will go with to original suggestion by @hanslovsky and change the javadoc.

gselzer:
@maarzt sorry for the delayed answer to your question, but yeah, I think this could be super useful in Ops. As scijava/scijava-ops#31 suggests we have no good multithreading solution as of this new iteration of ops; I hacked together the `thread.chunker` op to work for the time being but that was designed as a temporary fix. @ctrueden we could probably implement this across many (all?) of our lifting transformations, no?

src/main/java/net/imglib2/parallel/Parallelization.java
ctrueden: My initial thought is that using `runWithExecutor` in the `ModuleRunner` class, which governs the module execution process, would be sufficient for configuring SciJava module execution appropriately. For now, my vote is to make `setExecutorRequiresReset` private, but leave it otherwise exactly the same including the javadoc, since it is very helpful to understand what it is for. It would be fine to later make it public, should we encounter a case where that is needed, whereas making it private later is a breaking API change.

src/main/java/net/imglib2/parallel/ForkJoinExecutorService.java
maarzt: Now the differences between ForkJoinPool and ForkJoinExecutorService are explained in the javadoc.

src/main/java/net/imglib2/parallel/DefaultTaskExecutor.java
maarzt: I wouldn't change this for now. The purpose of `suggestNumberOfTasks` is to return a reasonable default value for parallelization. Something went wrong, if a value close to `Integer.MAX_VALUE` is returned. And I think an "Requested array size exceeds VM limit" exception will quickly lead to the correct conclusion: `suggestNumberOfTasks` and `getParallism` returned wrong values, that are way too big. The purpose of the line in the code is to prevent integer overflow, which will turn a number that is way to big into a negative number.

There should be a separate issue, that targets the problem that `Integer.MAX_VALUE` is not the upper bound for arrays.

ctrueden:
@maarzt @tpietzsch Can this PR please be resolved and merged soon? I need it for scijava-ops, as well as for pyimagej. This work hopefully solves #237.

PR #271, opened 2019-09-25
Fix Util.getArrayOrCellImgFactory to use Integer.MAX_VALUE - 8

maarzt:
The method Utils.getArrayOrCellImgFactory failed to return a proper
ImgFactory for images of the size Integer.MAX_VALUE. This is because
the actual maximum possible java array size, depends on the JVM and
is slightly below Integer.MAX_VALUE.

We use MAX_ARRAY_SIZE = Integer.MAX_VALUE - 8.
This is similar to OpenJDK8's ArrayList.

### DIFF
--- a/src/main/java/net/imglib2/util/Util.java
+++ b/src/main/java/net/imglib2/util/Util.java
@@ -71,6 +71,14 @@ import java.util.stream.StreamSupport;
  */
 public class Util
 {
+
+	/**
+	 * The possible java array size is JVM dependent an actually
+	 * slightly below Integer.MAX_VALUE. This is the same MAX_ARRAY_SIZE
+	 * as used for example in ArrayList in OpenJDK8.
+	 */
+	private static final int MAX_ARRAY_SIZE = Integer.MAX_VALUE - 8;
+
 	/**
 	 * This does only work when T is erased to Object at call site.
 	 * <p>
@@ -836,10 +844,7 @@ public class Util
 	 */
 	public static < T extends NativeType< T > > ImgFactory< T > getArrayOrCellImgFactory( final Dimensions targetSize, final T type )
 	{
-		if ( Intervals.numElements( targetSize ) <= Integer.MAX_VALUE )
-			return new ArrayImgFactory<>( type );
-		final int cellSize = ( int ) Math.pow( Integer.MAX_VALUE / type.getEntitiesPerPixel().getRatio(), 1.0 / targetSize.numDimensions() );
-		return new CellImgFactory<>( type, cellSize );
+		return getArrayOrCellImgFactory( targetSize, Integer.MAX_VALUE, type );
 	}
 
 	/**
@@ -860,13 +865,13 @@ public class Util
 	 */
 	public static < T extends NativeType< T > > ImgFactory< T > getArrayOrCellImgFactory( final Dimensions targetSize, final int targetCellSize, final T type )
 	{
-		if ( Intervals.numElements( targetSize ) <= Integer.MAX_VALUE )
+		Fraction entitiesPerPixel = type.getEntitiesPerPixel();
+		final long numElements = Intervals.numElements( targetSize );
+		final long numEntities = entitiesPerPixel.mulCeil( numElements );
+		if ( numElements <= Integer.MAX_VALUE && numEntities <= MAX_ARRAY_SIZE )
 			return new ArrayImgFactory<>( type );
-		final int cellSize;
-		if ( Math.pow( targetCellSize, targetSize.numDimensions() ) <= Integer.MAX_VALUE )
-			cellSize = targetCellSize;
-		else
-			cellSize = ( int ) Math.pow( Integer.MAX_VALUE / type.getEntitiesPerPixel().getRatio(), 1.0 / targetSize.numDimensions() );
+		final int maxCellSize = ( int ) Math.pow( Math.min( MAX_ARRAY_SIZE / entitiesPerPixel.getRatio(), Integer.MAX_VALUE ), 1.0 / targetSize.numDimensions() );
+		final int cellSize = Math.min( targetCellSize, maxCellSize );
 		return new CellImgFactory<>( type, cellSize );
 	}
 
--- a/src/test/java/net/imglib2/img/array/ArrayImgTest.java
+++ b/src/test/java/net/imglib2/img/array/ArrayImgTest.java
@@ -34,7 +34,14 @@
 
 package net.imglib2.img.array;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assume.assumeTrue;
+
+import net.imglib2.Cursor;
+import net.imglib2.img.Img;
+import net.imglib2.type.numeric.integer.Unsigned2BitType;
 import net.imglib2.type.numeric.real.FloatType;
 import net.imglib2.util.ImgTestHelper;
 import net.imglib2.util.Util;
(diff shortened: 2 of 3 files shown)


tpietzsch:
The tests on Travis error with
```
[ERROR] testGetArrayOrCellImgFactory(net.imglib2.util.UtilTest)  Time elapsed: 0.053 s  <<< ERROR!
373java.lang.OutOfMemoryError: Java heap space
374	at net.imglib2.util.UtilTest.testGetArrayOrCellImgFactory(UtilTest.java:240)
375
376[ERROR] testGetArrayOrCellImgFactoryWithCellSize(net.imglib2.util.UtilTest)  Time elapsed: 0.053 s  <<< ERROR!
377java.lang.OutOfMemoryError: Java heap space
```
Can you find a workaround? Like running tests only if enough heap space is available?

tpietzsch:
There are also probably a few more places that need fixing...
Here are places where Integer.MAX_VALUE appears in the current master:
```
~/workspace/imglib/imglib2$ find . -name "*.java" -exec grep -H -n "Integer.MAX_VALUE" {} \;
./src/test/java/net/imglib2/test/ImgLib2AssertTest.java:119:		ImgLib2Assert.assertImageEqualsIntegerType( img( Integer.MAX_VALUE ), img( Long.MAX_VALUE ) );
./src/test/java/net/imglib2/converter/RealTypeConverterInternalsTest.java:137:		// correctly by float. This tends to round up Integer.MAX_VALUE
./src/test/java/net/imglib2/loops/ClassCopyProviderBenchmark.java:88:			final IntType min = new IntType( Integer.MAX_VALUE );
./src/test/java/net/imglib2/histogram/HistogramNdTest.java:280:			v.set( ( int ) ( Math.random() * Integer.MAX_VALUE ) );
./src/main/java/net/imglib2/converter/RealTypeConverterInternals.java:115:		final boolean upperBound = variable.getMaxValue() <= Integer.MAX_VALUE;
./src/main/java/net/imglib2/util/Util.java:839:		if ( Intervals.numElements( targetSize ) <= Integer.MAX_VALUE )
./src/main/java/net/imglib2/util/Util.java:841:		final int cellSize = ( int ) Math.pow( Integer.MAX_VALUE / type.getEntitiesPerPixel().getRatio(), 1.0 / targetSize.numDimensions() );
./src/main/java/net/imglib2/util/Util.java:863:		if ( Intervals.numElements( targetSize ) <= Integer.MAX_VALUE )
./src/main/java/net/imglib2/util/Util.java:866:		if ( Math.pow( targetCellSize, targetSize.numDimensions() ) <= Integer.MAX_VALUE )
./src/main/java/net/imglib2/util/Util.java:869:			cellSize = ( int ) Math.pow( Integer.MAX_VALUE / type.getEntitiesPerPixel().getRatio(), 1.0 / targetSize.numDimensions() );
./src/main/java/net/imglib2/util/FlatCollections.java:405:		if ( size > Integer.MAX_VALUE )
./src/main/java/net/imglib2/img/array/ArrayImg.java:51: * number of basic types stored to {@link Integer#MAX_VALUE}. Keep in mind that
./src/main/java/net/imglib2/img/array/ArrayImg.java:68:	// TODO check for the size of numPixels being < Integer.MAX_VALUE?
./src/main/java/net/imglib2/img/array/ArrayImgFactory.java:101:		if ( numEntities > Integer.MAX_VALUE )
./src/main/java/net/imglib2/img/array/ArrayImgFactory.java:102:			throw new RuntimeException( "Number of elements in ArrayImg too big, use for example CellImg instead: " + numEntities + " > " + Integer.MAX_VALUE );
./src/main/java/net/imglib2/img/cell/CellImgFactory.java:120:	 * cell does not contain more than {@code Integer.MAX_VALUE} entities.
./src/main/java/net/imglib2/img/cell/CellImgFactory.java:136:		if ( numEntities > Integer.MAX_VALUE )
./src/main/java/net/imglib2/img/list/ListImg.java:47: * entities stored is limited to {@link Integer#MAX_VALUE}.
./src/main/java/net/imglib2/img/list/AbstractListImg.java:48: * stored is limited to {@link Integer#MAX_VALUE}.
./src/main/java/net/imglib2/type/AbstractBitType.java:50:	// Maximum count is Integer.MAX_VALUE * (64 / getBitsPerPixel())
./src/main/java/net/imglib2/type/logic/BitType.java:58:	// Maximum count is Integer.MAX_VALUE * (64 / getBitsPerPixel())
./src/main/java/net/imglib2/type/numeric/integer/IntType.java:138:		return Integer.MAX_VALUE;
./src/main/java/net/imglib2/type/numeric/integer/AbstractIntegerBitType.java:53:	// Maximum count is Integer.MAX_VALUE * (64 / getBitsPerPixel())
./src/main/java/net/imglib2/AbstractLocalizableInt.java:43: * limited to {@link Integer#MAX_VALUE} in every dimension.```

**Comment by ctrueden** [comment]
> Can you find a workaround? Like running tests only if enough heap space is available?

Ideally Travis would run it... do you know how much RAM the container has available?

Note that the pom-scijava-base configures maven-surefire-plugin's forked JVM(s) to run with 512MB by default.

If Travis does have more than 2GB, you could maybe split that specific test into its own file like `LargeMemoryIT.java` so that it runs with `maven-failsafe-plugin` instead of `maven-surefire-plugin`, and configure failsafe to use `<configuration><argLine>-Xmx2400m</argLine></configuration>` or however much is needed to successfully run the test. Then the surefire JVM(s) wouldn't all run with >2GB memory. But I suppose it isn't that important... either Travis has enough RAM or it doesn't, ultimately.

**Comment by ctrueden** [comment]
If no one has time to mess around further with the CI, then my vote here would be to simply not test this boundary condition. (❗️)

**Comment by maarzt** [comment]
I fixed the tests by adding a `assumeTrue` statement. This will skip the two tests, if there is less than 2 GB available.

**Comment by maarzt** [comment]
I looked into the other usages of `Integer.MAX_VALUE` in imglib2. I my opinion, they don't need to be fixed. They are responsible for giving nicer exception texts, when images or cells are to big. New edge cases would be introduced, if we lower the boundary to `Integer.MAX_VALUE - 8`. For example on my machine it's possible to create an ArrayImg with `Integer.MAX_VALUE - 2` bytes, this would no longer be possible.

**Comment by maarzt** [comment]
@tpietzsch, @ctrueden This should be ready to be merged.

File: src/main/java/net/imglib2/util/Util.java [hunk 1]
```diff
@@ -860,13 +868,13 @@ else if ( b == 1 )
 	 */
 	public static < T extends NativeType< T > > ImgFactory< T > getArrayOrCellImgFactory( final Dimensions targetSize, final int targetCellSize, final T type )
 	{
-		if ( Intervals.numElements( targetSize ) <= Integer.MAX_VALUE )
+		if ( Intervals.numElements( targetSize ) <= MAX_ARRAY_SIZE )
```
hanslovsky: When adjusting the `cellSize` below, the `MAX_ARRAY_SIZE` (or formerly `Integer.MAX_VALUE`) is divided by `type.getEntitiesPerPixel().getRatio()`. Wouldn't it make sense to also consider `type.getEntitiesPerPixel().getRatio()` in the checks for the number of required array elements for `ArrayImg` and cell size? I.e.
 1. 
```java
if ( Intervals.numElements( targetSize ) * type.getEntitiesPerPixel().getRatio() <= MAX_ARRAY_SIZE )
```
 2. 
```java
if ( Math.pow( targetCellSize, targetSize.numDimensions() ) * type.getEntitiesPerPixel().getRatio() <= MAX_ARRAY_SIZE )
```

This is not necessarily responsibility of this PR but could be fixed while at it.

maarzt: @hanslovsky Thank you for pointing that out. I changed the code accordingly.

PR #272, opened 2019-09-30
Add a simpler ImgView.wrap method

ctrueden:
This PR adds a new method that leans on `Util.getSuitableImgFactory` and `Util.getTypeFromInterval` to wrap the RAI with sensible defaults. It also adds basic unit tests for the `ImgView` class.

Wrapping a `RandomAccessibleInterval` to an `Img` previously required specifying an `ImgFactory`.

### DIFF
--- a/src/main/java/net/imglib2/img/ImgView.java
+++ b/src/main/java/net/imglib2/img/ImgView.java
@@ -51,8 +51,7 @@ import net.imglib2.view.iteration.SubIntervalIterable;
  * @author Tobias Pietzsch
  * @author Christian Dietz
  */
-public class ImgView< T extends Type< T > > extends
-		IterableRandomAccessibleInterval< T > implements Img< T >, SubIntervalIterable< T >
+public class ImgView< T extends Type< T > > extends IterableRandomAccessibleInterval< T > implements Img< T >, SubIntervalIterable< T >
 {
 
 	// factory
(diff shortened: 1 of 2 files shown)


src/main/java/net/imglib2/img/ImgView.java
```diff
@@ -154,25 +156,47 @@ public Object subIntervalIterationOrder( final Interval interval )
...
+	 *            RandomAccessibleInterval which will be wrapped with an ImgView
+	 * @return RandomAccessibleInterval represented as an Img
+	 * @see Util#getSuitableImgFactory(net.imglib2.Dimensions, Object)
 	 */
-	public static < T extends Type< T >> Img< T > wrap(final RandomAccessibleInterval< T > accessible, final ImgFactory< T > factory){
-		if(accessible instanceof Img){
-			return (Img<T>) accessible;
-		}else{
-			return new ImgView< T >(accessible, factory);
+	public static < T extends Type< T > > Img< T > wrap( final RandomAccessibleInterval< T > accessible )
```
hanslovsky: In order to avoid a breaking change, should we keep the existing singature for `wrap`? Maybe deprecate?

ctrueden: I did. And deprecated it. No?

hanslovsky: Oh yes, you did. I just didn't look carefully enough, my bad.

PR #277, opened 2019-12-12
Throw exception when trying to create Img with non-positive dimension

### LINKED ISSUE #276: ArrayImg with negative dimensions is possible
```java
Img<UnsignedByteType> img = ArrayImgs.unsignedBytes(-3, -2);
for (UnsignedByteType t : img)
	System.out.println(t);
```
prints:
```
0
0
0
0
0
0
```

But:
```java
Img<UnsignedByteType> img = ArrayImgs.unsignedBytes(-3, 2);
for (UnsignedByteType t : img)
	System.out.println(t);
```
prints:
```
Exception in thread "main" java.lang.NegativeArraySizeException
	at net.imglib2.img.basictypeaccess.array.AbstractByteArray.<init>(AbstractByteArray.java:50)
	at net.imglib2.img.basictypeaccess.array.ByteArray.<init>(ByteArray.java:46)
	at net.imglib2.img.basictypeaccess.array.ByteArray.createArray(ByteArray.java:57)
	at net.imglib2.img.basictypeaccess.array.ByteArray.createArray(ByteArray.java:42)
	at net.imglib2.img.array.ArrayImgFactory.create(ArrayImgFactory.java:91)
	at net.imglib2.img.array.ArrayImgFactory.create(ArrayImgFactory.java:68)
	at net.imglib2.img.array.ArrayImgs.unsignedBytes(ArrayImgs.java:107)
	at net.restlesscoder.sandbox.Sandbox.main(Sandbox.java:11)
```

Probably we should assert all lengths > 0 when creating new images.

hanslovsky:
Fixes #276

I tried to fix it not only in `ArrayImg` but in other `Img` implementations as well. I am currently enforcing all dimensions to be positive, but zero-size dimensions may make sense, as well.

cc @ctrueden

### DIFF
--- a/src/main/java/net/imglib2/Dimensions.java
+++ b/src/main/java/net/imglib2/Dimensions.java
@@ -98,6 +98,18 @@ public interface Dimensions extends EuclideanSpace
 		return dimensions;
 	}
 
+	/**
+	 * Verify that {@code dimensions} is not null or empty, and that all
+	 * dimensions are positive. Throw {@link InvalidDimensions} otherwise.
+	 *
+	 * @param dimensions
+	 *            to be verified.
+	 * @return {@code dimensions} if successfully verified.
+	 * @throws IllegalArgumentException
+	 *             if {@code dimensions == null} or
+	 *             {@code dimensions.length == 0} or any dimensions is zero or
+	 *             negative.
+	 */
 	public static long[] verify( final long... dimensions ) throws InvalidDimensions
 	{
 		if ( dimensions == null )
--- a/src/main/java/net/imglib2/Dimensions.java
+++ b/src/main/java/net/imglib2/Dimensions.java
@@ -66,6 +66,14 @@ public interface Dimensions extends EuclideanSpace
 	 */
 	public long dimension( int d );
 
+	/*
+	 * -----------------------------------------------------------------------
+	 *
+	 * Static methods
+	 *
+	 * -----------------------------------------------------------------------
+	 */
+
 	/**
 	 * Check whether all entries in {@code dimensions} are positive
 	 *
@@ -73,9 +81,24 @@ public interface Dimensions extends EuclideanSpace
 	 * @return true if all entries in {@code dimension} are positive, false
 	 *         otherwise
 	 */
-	public static boolean allPositive( final long... dimensions )
+	static boolean allPositive( final long... dimensions )
 	{
 		for ( final long d : dimensions )
+			if ( d < 1 )
+				return false;
+		return true;
+	}
+
+	/**
+	 * Check whether all entries in {@code dimensions} are positive
+	 *
+	 * @param dimensions
+	 * @return true if all entries in {@code dimension} are positive, false
+	 *         otherwise
+	 */
+	static boolean allPositive( final int... dimensions )
+	{
+		for ( final int d : dimensions )
 			if ( d < 1 )
 				return false;
 		return true;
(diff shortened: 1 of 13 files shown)


bogovicj:
> zero-size dimensions may make sense

Are you thinking something like dims size `[2,2,2,0]`?

hanslovsky:
> Are you thinking something like dims size `[2,2,2,0]`?

Yes, it may make sense to allow for that (and `Img` or `Dimensions` could even have an `isEmpty`) .

src/main/java/net/imglib2/img/cell/CellImgFactory.java
```diff
@@ -164,6 +164,7 @@ public static void verifyDimensions( final long dimensions[] ) throws IllegalArg
 			final T type,
 			final NativeTypeFactory< T, A > typeFactory )
 	{
+		Dimensions.InvalidDimensions.checkAllPositiveOrElseThrow( dimensions );
```
ctrueden: The subsequent `verifyDimensions` check is already checking for positive dimensions, and throwing an `IllegalArgumentException` if not.

src/main/java/net/imglib2/Dimensions.java
```diff
@@ -61,4 +64,42 @@ default void dimensions( final long[] dimensions )
...
 	 */
 	public long dimension( int d );
+
+	/**
+	 * Check that all entries in dimensions are positive
+	 *
+	 * @param dimensions
+	 * @return true if all entries in dimension are positive, false otherwise
+	 */
+	public static boolean dimensionsCheckAllPositive( final long[] dimensions )
```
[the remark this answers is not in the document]
hanslovsky: I will rename to`allPositive`, it reads better.

I am conflicted about `long...`. I had used it initially, but it is only useful when you check dimensions by hand, and I don't think that something like this is very common:
```
Dimensions.allPositive(1, 2, 3, ....);
ArrayImgs.longs(1, 2, 3, ...);
```
Instead of typing the dimensions twice, I would rather imagine something like this:
```
final long[] dims = {1, 2, 3, ...};
Dimesnions.allPositive(dims);
ArrayImgs.longs(dims);
```

Maybe this could be addressed by changing the return type to `long[]` and returning the argument if the check passes (`null` otherwise or throw exception).

src/main/java/net/imglib2/img/cell/CellImgFactory.java
hanslovsky: I will look into it.

hanslovsky:
> Since we don't use custom RuntimeException extensions in other scenarios, it feels weird to me to introduce one here—it makes me want to consider doing so in more places, for better type safety. My weak preference would be for (A) above and ditch the custom exception type.

I think I would favor (B) because it is still consistent with previous behavior but still contains the new information.

> (Although to get really general, we may want to extend `Exception` rather than `RuntimeException`...)

At the cost of adding checked exception (i.e. breaking API), though. I don't think we should go that far.

hanslovsky:
Regarding of verifying positivity (`> 0`) or non-negativity (`>= 0`), the `CellImgFactory` [verifies positivity](https://github.com/imglib/imglib2/blob/617c5e515c4ce3285ce3bd3d29b25b8e51ed5ece/src/main/java/net/imglib2/img/cell/CellImgFactory.java#L112-L114):
```java
for ( int d = 0; d < dimensions.length; d++ )
	if ( dimensions[ d ] <= 0 )
		throw new IllegalArgumentException( "dimensions[ " + d + " ] <= 0" );
```

We could just follow this convention and require positive dimensions (which is the current state).

src/main/java/net/imglib2/img/cell/CellImgFactory.java
hanslovsky: I moved the call to `Dimensions.verifyAllPositive` into `CellImgFactory.verifyDimensions`. We could even extract `verifyDimensions` into the `Dimensions` class.

tpietzsch:
Could you add javadoc for `Dimensions.verify` please (can be adapted from the deprecated `CellImgFactory` method)? Then I think it's ready to be merged.

hanslovsky:
Added the JavaDoc in 7d876ee

PR #278, opened 2019-12-15
Buffer-backed accesses

### LINKED ISSUE #5: Port NIO-backed images to ImgLib2
In ImgLib1, there is support for NIO-backed primitive arrays in the mpicbg.imglib.container.basictypecontainer.array package, specifically the NIO\* classes. This was never translated to ImgLib2. It would be nice to do so, since NIO-backed arrays have [a few advantages](http://onjava.com/pub/a/onjava/2002/10/02/javanio.html), including allocation outside the heap, arrays larger than 2GB, and shared memory with native applications.

hanslovsky:
This is very experimental draft PR to figure out the best interface for implementing buffer-backed accesses that were requested a long time ago (#5, the only remaining open single digit issue), and has come up at the Python/ImgLib2 shared memory hackathon in Dresden in Dec 2019. Please use this PR to discuss the exact design before going ahead and implementing all accesses (`byte`, `short`, `int`,...)

There are multiple open questions that I came up with when creating this first draft (see `TODO`s in the `LongBufferLongAccess` class):

 - General
  1. Implementing ArrayDataAccess probably makes sense
  2. Should we have separate accesses that implement VolatileAccess?
  3. Should we use ByteBuffer only or use specific buffers like LongBuffer. allocateDirect is only available in ByteBuffer. Both accesses could be implemented, i.e. LongBufferLongAccess as well as ByteBufferLongAccess
 - Specific to `LongBufferLongAccess.createArray`
  1. Should we respect the kind of buffer that this.buffer is? i.e. if this.buffer is direct, then the buffer of the new access should also be direct? Is there a method method in Buffer that allows to do that?

Fixes #5

### DIFF
--- a/src/main/java/net/imglib2/img/basictypeaccess/buffer/LongBufferLongAccess.java
+++ b/src/main/java/net/imglib2/img/basictypeaccess/buffer/LongBufferLongAccess.java
@@ -1,0 +1,100 @@
+package net.imglib2.img.basictypeaccess.buffer;
+
+import java.nio.ByteBuffer;
+import java.nio.LongBuffer;
+
+import net.imglib2.img.basictypeaccess.LongAccess;
+import net.imglib2.img.basictypeaccess.array.ArrayDataAccess;
+import net.imglib2.img.basictypeaccess.volatiles.VolatileAccess;
+
+public class LongBufferLongAccess implements ArrayDataAccess< LongBufferLongAccess >, LongAccess, VolatileAccess
+{
+
+	// TODO How to properly design a buffer access?
+	// 1. Implementing ArrayDataAccess probably makes sense
+	// 2. Should we have separate accesses that implement VolatileAccess?
+	// 3. Should we use ByteBuffer only or use specific buffers like LongBuffer.
+	//    allocateDirect is only available in ByteBuffer.
+	//    Both accesses could be implemented, i.e. LongBufferLongAccess as well
+	//    as ByteBufferLongAccess
+
+	private static boolean DEFAULT_IS_VALID = true;
+
+	private final LongBuffer buffer;
+
+	private final boolean isValid;
+
+	public LongBufferLongAccess( final int numEntities )
+	{
+		this( numEntities, DEFAULT_IS_VALID );
+	}
+
+	public LongBufferLongAccess( final int numEntities, final boolean isValid )
+	{
+		this( LongBuffer.allocate( numEntities ), isValid );
+	}
+
+	public LongBufferLongAccess( final ByteBuffer buffer )
+	{
+		this( buffer, DEFAULT_IS_VALID );
+	}
+
+	public LongBufferLongAccess( final ByteBuffer buffer, final boolean isValid )
+	{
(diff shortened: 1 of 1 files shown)


tpietzsch:
> 3\. Should we use ByteBuffer only or use specific buffers like LongBuffer. allocateDirect is only available in ByteBuffer. Both accesses could be implemented, i.e. LongBufferLongAccess as well as ByteBufferLongAccess

You can allocate a direct ByteBuffer and then have a `asLongBuffer()` view on it.

hanslovsky:
`LongBufferLongAccess.createArray` now creates a direct buffer, if the internal buffer is direct.

If we use the typed buffers (i.e. `LongBuffer` etc) we could also go with a more simple naming scheme (`LongBuffer` etc).

hanslovsky:
@axtimwalde This is probably relevant for avoiding copies in the N5 `CacheLoader`.

hanslovsky:
re:
> Should we use ByteBuffer only or use specific buffers like LongBuffer.

I think we should use specific buffers because then we don't need to consider byte sizes of the individual types, e.g.:
```kotlin
// create a byte buffer from numpy array through kotlin-numpy
val access = (arange(15.0)+10.0).data!!
// try to get the second double but really getting some double "in between"
println(access.getDouble(1))
// get second double
println(access.asDoubleBuffer().get(1))
// get second double
println(access.getDouble(8))
```
In the 
output:
```
1.79570413737339E-307
11.0
11.0
```

mkitti:
I'm considering adopting this PR. It would be useful to use together with [com.sun.jna.Pointer.getByteBuffer](https://java-native-access.github.io/jna/4.2.1/com/sun/jna/Pointer.html#getByteBuffer-long-long-) to avoid excess copying when dealing with native libraries.

src/main/java/net/imglib2/img/basictypeaccess/buffer/LongBufferLongAccess.java
```diff
@@ -0,0 +1,104 @@
...
+	@Override
+	public LongBuffer getCurrentStorageArray()
+	{
+		return buffer;
+	}
+
+	@Override
+	public int getArrayLength()
+	{
+		return buffer.capacity();
```
mkitti, line 101:
```suggestion
		return buffer.limit();
```

`getArrayLength()` should be `limit()` because that defines practical length of the data in `Buffer`.

https://docs.oracle.com/javase/8/docs/api/java/nio/Buffer.html

> A buffer's capacity is the number of elements it contains. The capacity of a buffer is never negative and never changes.
> 
> A buffer's limit is the index of the first element that should not be read or written. A buffer's limit is never negative and is never greater than its capacity.

`buffer.get(int index)` throws a `IndexOutOfBoundsException` if `index >= buffer.limit()`
https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#get-int-

> IndexOutOfBoundsException - If index is negative or not smaller than the buffer's limit

hanslovsky:
Close in favor of #299

PR #280, opened 2020-01-24
Re-add type-bounded signatures of Views.extendValue and Views.expandValue

imagejan:
See [this discussion](https://gitter.im/imglib/imglib2?at=5e2abf31e177666195bd4689) on gitter.

### DIFF
--- a/src/main/java/net/imglib2/view/Views.java
+++ b/src/main/java/net/imglib2/view/Views.java
@@ -65,6 +65,7 @@ import net.imglib2.transform.integer.permutation.SingleDimensionPermutationTrans
 import net.imglib2.transform.integer.shear.InverseShearTransform;
 import net.imglib2.transform.integer.shear.ShearTransform;
 import net.imglib2.type.BooleanType;
+import net.imglib2.type.Type;
 import net.imglib2.type.numeric.IntegerType;
 import net.imglib2.type.numeric.NumericType;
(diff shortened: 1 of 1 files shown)


src/main/java/net/imglib2/view/Views.java
```diff
@@ -1356,6 +1375,24 @@ public static boolean isZeroMin( final Interval interval )
...
+	 * @param source
+	 *            the interval to expand.
+	 * @param t
+	 *            Constant extension of source.
+	 * @return Expansion of the {@link RandomAccessibleInterval} source as
+	 *         specified by t and border.
+	 * @deprecated use {@code expandValue} with unbounded type parameter T
+	 */
+	@Deprecated
+	public static < T extends RealType< T > > IntervalView< T > expandValue( final RandomAccessibleInterval< T > source, final T t, final long... border )
```
hanslovsky: This can be relaxed to `Type<T>` instead of `RealType<T>`, no?

tpietzsch: Yes, it should be as it was before #268

imagejan: Oops, you're right, I guess..

imagejan: Now changed and force-pushed.

PR #281, opened 2020-01-28
Re-add equalDimensions signature with Interval

imagejan:
The signature was relaxed from `Interval` arguments to `Dimensions` arguments in commit 512423f. This commit re-introduces an `Interval`-based signature to avoid binary API breakage.

### DIFF
--- a/src/main/java/net/imglib2/util/Intervals.java
+++ b/src/main/java/net/imglib2/util/Intervals.java
@@ -846,7 +846,7 @@ public class Intervals
 	}
 
 	/**
-	 * Tests whether two intervals have equal dimensions (same size).
+	 * Tests whether two {@link Dimensions} have the same size.
 	 */
 	public static boolean equalDimensions( final Dimensions a, final Dimensions b )
 	{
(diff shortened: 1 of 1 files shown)


src/main/java/net/imglib2/util/Intervals.java
```diff
@@ -860,6 +860,14 @@ public static boolean equalDimensions( final Dimensions a, final Dimensions b )
 		return true;
 	}
 
+	/**
+	 * Tests whether two intervals have equal dimensions (same size).
+	 */
+	public static boolean equalDimensions( final Interval a, final Interval b )
+	{
+		return equalDimensions((Dimensions) a, (Dimensions) b);
```
ctrueden: Tiny style formatting issue with this line: parentheses lack surrounding whitespace.

ctrueden:
LGTM!

src/main/java/net/imglib2/util/Intervals.java
imagejan:
```suggestion
		return equalDimensions( ( Dimensions ) a, ( Dimensions ) b );
```

PR #287, opened 2020-07-16
Numeric implements Pow, PowFloatingPoint

acardona:
This request is a starting point for an implementation of pow in imglib2.

Note it remains unimplemented for ```Unsigned128BitType```.

### DIFF
--- a/src/main/java/net/imglib2/type/numeric/integer/AbstractIntegerBitType.java
+++ b/src/main/java/net/imglib2/type/numeric/integer/AbstractIntegerBitType.java
@@ -140,6 +140,12 @@ public abstract class AbstractIntegerBitType< T extends AbstractIntegerBitType< 
 
 	@Override
 	public void div(final T t) { set(get() / t.get()); }
+	
+	@Override
+	public void pow(final T t) { set( (long) Math.pow( get(), t.get() ) ); }
+	
+	@Override
+	public void pow(final double d) { set( (long) Math.pow( get(), d) ); }
 
 	@Override
 	public void set( final T c ) { set( c.get() ); }
--- a/src/main/java/net/imglib2/type/numeric/integer/AbstractIntegerBitType.java
+++ b/src/main/java/net/imglib2/type/numeric/integer/AbstractIntegerBitType.java
@@ -142,10 +142,10 @@ public abstract class AbstractIntegerBitType< T extends AbstractIntegerBitType< 
 	public void div(final T t) { set(get() / t.get()); }
 	
 	@Override
-	public void pow(final T t) { set( (long) Math.pow( get(), t.get() ) ); }
+	public void pow(final T t) { setReal( Math.pow( get(), t.get() ) ); }
 	
 	@Override
-	public void pow(final double d) { set( (long) Math.pow( get(), d) ); }
+	public void pow(final double power) { setReal( Math.pow( get(), power) ); }
 
 	@Override
 	public void set( final T c ) { set( c.get() ); }
--- a/src/main/java/net/imglib2/type/numeric/integer/GenericIntType.java
+++ b/src/main/java/net/imglib2/type/numeric/integer/GenericIntType.java
@@ -181,14 +181,14 @@ public abstract class GenericIntType< T extends GenericIntType< T > > extends Ab
 	public void pow( final T c )
 	{
 		final int a = getInt();
-		setInt( ( int )Math.pow( a, c.getInt() ) );
+		setReal( Math.pow( a, c.getInt() ) );
 	}
 	
 	@Override
 	public void pow( final double power )
 	{
 		final int a = getInt();
-		setInt( ( int )Math.pow( a, power ) );
+		setReal( Math.pow( a, power ) );
 	}
 
 	@Override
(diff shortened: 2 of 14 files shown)


src/main/java/net/imglib2/type/operators/PowFloatingPoint.java
```diff
@@ -0,0 +1,44 @@
+/*
+ * #%L
+ * ImgLib2: a general-purpose, multidimensional image processing library.
+ * %%
+ * Copyright (C) 2009 - 2018 Tobias Pietzsch, Stephan Preibisch, Stephan Saalfeld,
```
imagejan, line 5: Running `mvn license:update-file-header` should automatically update this to 2020.

src/main/java/net/imglib2/type/operators/PowFloatingPoint.java
```diff
@@ -0,0 +1,44 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ * #L%
+ */
+package net.imglib2.type.operators;
+
+/**
+ * @author Albert Cardona
+ *
+ * @param <T>
```
imagejan, line 39: There's no parameter `T` in this interface.

acardona: Thanks, fixed that, rebased.

src/main/java/net/imglib2/type/operators/PowFloatingPoint.java
acardona: Thanks, fixed, rebased.

src/main/java/net/imglib2/type/numeric/integer/UnsignedVariableBitLengthType.java
```diff
@@ -280,6 +280,16 @@ public void mul( final UnsignedVariableBitLengthType t )
 		set( get() * t.get() );
 	}
 
+	@Override
+	public void pow( final UnsignedVariableBitLengthType c) {
+		set( ( long )Math.pow( get(), c.get() ) );
```
axtimwalde, line 285: Same question as above, shouldn't we leave the casting to the `setReal(double)` method? I would expect rounding.

src/main/java/net/imglib2/type/numeric/integer/AbstractIntegerBitType.java
```diff
@@ -140,6 +140,12 @@ public AbstractIntegerBitType(
 
 	@Override
 	public void div(final T t) { set(get() / t.get()); }
+	
+	@Override
+	public void pow(final T t) { set( (long) Math.pow( get(), t.get() ) ); }
```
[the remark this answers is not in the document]
acardona: I'd be happy to use setReal, makes more sense as it enables subclasses to modify the behavior.

src/main/java/net/imglib2/type/numeric/integer/GenericIntType.java
```diff
@@ -176,6 +176,20 @@ public void sub( final T c )
...
+	{
+		final int a = getInt();
+		setInt( ( int )Math.pow( a, c.getInt() ) );
+	}
+	
+	@Override
+	public void pow( final double power )
+	{
+		final int a = getInt();
+		setInt( ( int )Math.pow( a, power ) );
```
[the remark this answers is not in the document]
acardona: Same: didn't think about it. Casting is an implicit floor. setReal makes sense and enables subclasses to change the behavior.

src/main/java/net/imglib2/type/numeric/integer/UnsignedVariableBitLengthType.java
acardona: Yes, agreed.

acardona:
Any comments @axtimwalde @tpietzsch on the unimplemented ```Unsigned128BitType.pow``` methods? I'm happy to leave them as they are.

axtimwalde:
No, I think leaving them unimplemented is ok for now.  I am personally more interested in an integer version of pow for which there is an easy to replicate algorithm in Guava https://github.com/google/guava/blob/master/guava/src/com/google/common/math/IntMath.java#L227 but that is not covered in this pull request.  You could use this to implement the the `Pow` interface (that could also be interesting for all the other integer types like `ARGBType`, `LongType`..., but not `PowFloatingPoint`.

acardona:
Thanks @axtimwalde. On Guava and its ```pow``` method for integers: it's an Apache license, so we'll have to derive our integer implementation from first principles AKA from a mathematical formulation.

ctrueden:
> it's an Apache license, so we'll have to derive our integer implementation from first principles

I don't understand—Apache 2 is a permissive license. And it's compatible with GPL v3, as well as BSD/MIT/ISC-style licenses. The only potential issue is the GPLv2, right?

We are using Guava as a dependency for SciJava Ops, so if there is a licensing issue lurking there, I'd like to understand it. If we can't use Guava, it's going to be a big problem for the Ops matcher, since we rely heavily on Guava's TypeToken.

axtimwalde:
@acardona I concur with @ctrueden and canot see a problem, Apache 2 is super permissive as far as I understand.  Also, the first order principles in this implementation aren't particularly deep: Use bitshifts for powers of 2, loop over multiplications for the rest?

PR #288, opened 2020-07-21
Biconverter

axtimwalde:
This branch implements simple two valued pixel algebra and type conversion similar to type conversion with converters.  There are multiple solutions to do this potentially more efficient in loops or for blocks in dedicated ops, but the pixel-wise lazy version was missing as far as I know.  This is the opposite of what one would like to do in interpreted scripts, it's a convenience layer for compiled languages for the JVM, particularly those that support lambdas.  Adding two double images/ iterables/ functions fa, and fb looks like:
```
sum = Converters.convert(fa, fb, (a, b, c) -> c.set(a.get() + b.get()), new DoubleType());
```


axtimwalde:
I just realized that both Converters and BiConverters currently do not support stateful functions in a way that is safe for multi-threading.  This is something for another pull request.

axtimwalde:
Instead of doing another pull request I added supplier based constructors which allows us to use stateful Converters or BiConverters in multi-threaded code while re-using Converters and BiConverter instances when used as before (assuming that
they are stateless).

Issue #285, opened 2020-03-26
Add methods/constructors taking (Real)Localizables where relevant


bogovicj:
What do we think about adding methods/constructors taking `Localizables` or `RealLocalizables` where they currently take only `long[]` or `double[]` inputs?

E.g.
```
RealPoint minPt
RealPoint maxPt;
new FinalRealInterval( minPt, maxPt );
```

This could get annoying if we also wanted to support all of:
```
double[] min;
double[] max;
new FinalRealInterval( minPt, max );
new FinalRealInterval( min, maxPt );
```
but may not actually be so bad.

Related: a `Localizables.asDoubleArray( RealLocalizable )` method similar to [`asLongArray`](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/util/Localizables.java#L51) would help if we don't want to go this route.

This came up [on gitter](https://gitter.im/imglib/imglib2?at=5e7cbc5104405b1a718e6a80), in the context of a `net.imglib2.realtransform.Translation3D`
@imagejan 
@hanslovsky 
@igorpisarev

tpietzsch:
> What do we think about adding methods/constructors taking `Localizables` or `RealLocalizables` where they currently take only `long[]` or `double[]` inputs?

Would be nice, yes.

> new FinalRealInterval( minPt, max );

I wouldn't add those proactively...


I'm definitely for `Localizables.asDoubleArray()`.
It might be also a good idea to put these as static methods into the respective interfaces, i.e.,
`static long[] asLongArray( Localizable )` could go into `Localizable` and
`static double[] asDoubleArray( RealLocalizable )` could go into `RealLocalizable`.
(I would however not add non-static default methods `default long[] asLongArray()` to Localizable etc, because that might be pretty confusing in derived classes, e.g., `RandomAccess.asLongArray()`.)

hanslovsky:
> I wouldn't add those proactively...

I agree with that

> (I would however not add non-static default methods `default long[] asLongArray()` to `Localizable` etc, because that might be pretty confusing in derived classes, e.g., `RandomAccess.asLongArray()`.)

That makes total sense. What if those default methods would be named differently? e.g.
```
double[] RealLocalizable.localizeAsDoubleArray();
long[] Localizable.localizeAsLongArray();
```
and their 32bit counter-parts.

tpietzsch:
> What if those default methods would be named differently?

I still don't have a 100% good feeling with that. Maybe.

[...]

PR #289, opened 2020-07-21
add methods and constructors taking Localizable and RealLocalizable

bogovicj:
Changes are in line with the discussion in issue #285.  Two remaining points / questions:

I added `Localizable.asLongArray` and `RealLocalizable.asDoubleArray`which duplicate `Localizables.asLongArray` and `Localizables.asDoubleArray`.  I modified the `Localizables` methods to call the other methods as appropriate. Is this situation okay? Should we deprecate the `Localizables` methods?  Or better to leave the methods in `Localizables` and remove from `RealLocalizable` and `Localizable`?

I also added a few static methods in Intervals: 
* `Point Intervals.minAsPoint( Interval )`
* `Point Intervals.maxAsPoint( Interval )`
* `RealPoint Intervals.minAsRealPoint( RealInterval )`
* `RealPoint Intervals.maxAsRealPoint( RealInterval )`

not sure how useful those will be, but seem reasonable, and is in-line with the idea of this PR.

### DIFF
--- a/src/main/java/net/imglib2/Interval.java
+++ b/src/main/java/net/imglib2/Interval.java
@@ -191,26 +191,4 @@ public interface Interval extends RealInterval, Dimensions
 		return max;
 	}
 
-	/**
-	 * Allocates a new long array with the dimensions of this Interval.
-	 * 
-	 * @return the dimensions
-	 */
-	default long[] dimensionsAsLongArray()
-	{
-		long[] dims = new long[ numDimensions() ];
-		dimensions( dims );
-		return dims;
-	}
-
-	/**
-	 * Allocates a new {@link Point} with the dimensions of this Interval.
-	 * 
-	 * @return the dimensions
-	 */
-	default Point dimensionsAsPoint()
-	{
-		return new Point( dimensionsAsLongArray() );
-	}
-
 }
--- a/src/main/java/net/imglib2/Interval.java
+++ b/src/main/java/net/imglib2/Interval.java
@@ -165,8 +165,8 @@ public interface Interval extends RealInterval, Dimensions
 	 *
 	 * Please note that his method allocates a new {@link Point} each time
 	 * which introduces notable overhead in both compute and memory.
-	 * If you query it frequently, you should allocate a dedicated array
-	 * first and reuse it with {@link #min(long[])}.
+	 * If you query it frequently, you should allocate a dedicated
+	 * {@link Point} first and reuse it with {@link #min(Positionable)}.
 	 *
 	 * @return the min
 	 */
@@ -199,8 +199,8 @@ public interface Interval extends RealInterval, Dimensions
 	 *
 	 * Please note that his method allocates a new {@link Point} each time
 	 * which introduces notable overhead in both compute and memory.
-	 * If you query it frequently, you should allocate a dedicated array
-	 * first and reuse it with {@link #max(long[])}.
+	 * If you query it frequently, you should allocate a dedicated
+	 * {@link Point} first and reuse it with {@link #max(Positionable)}.
 	 *
 	 * @return the max
 	 */
--- a/src/main/java/net/imglib2/Interval.java
+++ b/src/main/java/net/imglib2/Interval.java
@@ -142,4 +142,75 @@ public interface Interval extends RealInterval, Dimensions
 	{
 		return max( d ) - min( d ) + 1;
 	}
+
+	/**
+	 * Allocates a new long array with the minimum of this Interval.
+	 * 
+	 * @return the min
+	 */
+	default long[] minAsLongArray()
+	{
+		long[] min = new long[ numDimensions() ];
+		min( min );
+		return min;
+	}
+
+	/**
+	 * Allocates a new {@link Point} with the maximum of this Interval.
+	 * 
+	 * @return the min
+	 */
+	default Point minAsPoint()
+	{
+		Point min = new Point( numDimensions() );
+		min( min );
+		return min;
+	}
+
+	/**
+	 * Allocates a new long array with the maximum of this Interval.
+	 * 
+	 * @return the max
+	 */
+	default long[] maxAsLongArray()
+	{
+		long[] max = new long[ numDimensions() ];
+		max( max );
+		return max;
+	}
+
+	/**
+	 * Allocates a new {@link Point} with the maximum of this Interval.
+	 * 
+	 * @return the max
+	 */
+	default Point maxAsPoint()
+	{
+		Point max = new Point( numDimensions() );
+		max( max );
+		return max;
+	}
+
+	/**
+	 * Allocates a new long array with the dimensions of this Interval.
+	 * 
+	 * @return the dimensions
+	 */
+	default long[] dimensionsAsLongArray()
+	{
+		long[] dims = new long[ numDimensions() ];
+		dimensions( dims );
+		return dims;
+	}
+
+	/**
+	 * Allocates a new {@link Point} with the dimensions of this Interval.
+	 * 
+	 * @return the dimensions
+	 */
+	default Point dimensionsAsPoint()
+	{
+		return new Point( dimensionsAsLongArray() );
+	}
+
 }
--- a/src/main/java/net/imglib2/Interval.java
+++ b/src/main/java/net/imglib2/Interval.java
@@ -11,13 +11,13 @@
  * %%
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
- * 
+ *
  * 1. Redistributions of source code must retain the above copyright notice,
  *    this list of conditions and the following disclaimer.
  * 2. Redistributions in binary form must reproduce the above copyright notice,
  *    this list of conditions and the following disclaimer in the documentation
  *    and/or other materials provided with the distribution.
- * 
+ *
  * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
  * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
  * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
@@ -40,7 +40,7 @@ package net.imglib2;
  * <em>x<sub>d</sub></em>&le;<em>max<sub>d</sub></em>;<em>d</em>&isin;{0&hellip;
  * <em>n</em>-1}}
  * </p>
- * 
+ *
  * <p>
  * An {@link Interval} over the discrete source domain. <em>Note</em> that this
  * does <em>not</em> imply that for <em>all</em> coordinates in the
@@ -50,7 +50,7 @@ package net.imglib2;
  * coordinate for each. By that, minimum and maximum are defined but the
  * {@link Interval} does not define a value for all coordinates in between.
  * </p>
- * 
+ *
  * @author Stephan Saalfeld
  * @author Stephan Preibisch
  */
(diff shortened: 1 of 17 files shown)


axtimwalde:
Not sure about the static methods.  I would be in favor of default implementations in the `Localizable`, `Interval` and `Dimensions` interfaces, like
```java
public default long[] minAsLongArray() {
   final long[] min = new long[numDimensions()];
   min(min);
}
[...]
public default long[] locationAsLongArray() {
   final long[] location = new long[numDimensions()];
   min(location);
}
[...]
```
Does that have any downsides?

bogovicj:
About default methods, in the discussion around this issue (https://github.com/imglib/imglib2/issues/285)  @tpietzsch said:
>(I would however not add non-static default methods default long[] asLongArray() to Localizable etc, because that might be pretty confusing in derived classes, e.g., RandomAccess.asLongArray().)

axtimwalde:
That would be confusing, but what if the names were non-confusing?
```
long[] locationToLongArray();
double[] locationToDoubleArray();
Point locationToPoint();

long[] minToLongArray();
double[] minToDoubleArray();
Point minToPoint();

long[] dimensionsToLongArray();
double[] dimensionsToDoubleArray();
Point dimensionsToPoint();

[...]
```

bogovicj:
I'd be happy with that.

bogovicj:
Updated this PR with default convenience methods with descriptive names.

src/main/java/net/imglib2/Localizable.java
```diff
@@ -84,6 +84,29 @@ default int getIntPosition( final int d )
 		return (int) getLongPosition( d );
 	}
 
+	/**
+	 * Allocate and return a long array containing the localizable's position.
+	 * 
+	 * @return the position
+	 */
+	default long[] locationToLongArray()
```
axtimwalde, line 92: May be positionAsLongArray?

src/main/java/net/imglib2/Localizable.java
```diff
@@ -84,6 +84,29 @@ default int getIntPosition( final int d )
...
+		return out;
+	}
+
+	/**
+	 * Allocate and return a {@link Point} containing the localizable's
+	 * position.
+	 * 
+	 * @return the position
+	 */
+	default Point locationToPoint()
```
axtimwalde, line 105: may be positionAsPoint?

src/main/java/net/imglib2/Localizable.java
```diff
@@ -104,4 +127,18 @@ default double getDoublePosition( final int d )
...
 	}
+
+	/**
+	 * Allocate and return a long array containing the localizable's position.
+	 * 
+	 * @param localizable
+	 *            the localizable
+	 * @return the array
+	 */
+	public static long[] asLongArray( final Localizable localizable )
```
axtimwalde, line 138: May be remove because we now have the defaults who do the same?

src/main/java/net/imglib2/RealLocalizable.java
```diff
@@ -70,6 +70,28 @@ default void localize( final double[] position )
 			position[ d ] = getDoublePosition( d );
 	}
 
+	/**
+	 * Allocate and return a double array with the position.
+	 * 
+	 * @return the position
+	 */
+	default double[] locationToDoubleArray()
```
axtimwalde, line 78: may be positionAsDoubleArray?

src/main/java/net/imglib2/RealLocalizable.java
```diff
@@ -70,6 +70,28 @@ default void localize( final double[] position )
...
+		localize( out );
+		return out;	
+	}
+
+	/**
+	 * Allocate and return a {@link RealPoint} with the current position.
+	 * 
+	 * @return the position
+	 */
+	default RealPoint locationToRealPoint()
```
axtimwalde, line 90: may be positionAsRealPoint?

src/main/java/net/imglib2/util/Intervals.java
```diff
@@ -954,6 +956,26 @@ public static boolean equalDimensions( final Interval a, final Interval b )
...
+	 * Keep in mind that creating arrays wildly is not good practice and
+	 * consider using the interval directly.
+	 * </p>
+	 * 
+	 * @param interval
+	 *            something with interval boundaries
+	 * 
+	 * @return minimum as a new Point
+	 */
+	public static Point minAsPoint( final Interval interval )
```
axtimwalde, line 972: maybe be remove because we now have the defaults

src/main/java/net/imglib2/util/Intervals.java
```diff
@@ -996,6 +1018,26 @@ public static boolean equalDimensions( final Interval a, final Interval b )
...
+	 * Keep in mind that creating arrays wildly is not good practice and
+	 * consider using the interval directly.
+	 * </p>
+	 * 
+	 * @param interval
+	 *            something with interval boundaries
+	 * 
+	 * @return maximum as a new Point
+	 */
+	public static Point maxAsPoint( final Interval interval )
```
axtimwalde, line 1034: may be remove because we now have the defaults

src/main/java/net/imglib2/util/Intervals.java
```diff
@@ -1018,6 +1060,28 @@ public static boolean equalDimensions( final Interval a, final Interval b )
...
+	 * consider using the interval directly. See
+	 * {@link RealInterval#realMin(double[])}.
+	 * </p>
+	 * 
+	 * @param interval
+	 *            something with interval boundaries
+	 * 
+	 * @return maximum as a new RealPoint 
+	 */
+	public static RealPoint maxAsRealPoint( final RealInterval interval )
```
axtimwalde, line 1078: may be remove because now we have the defaults

src/main/java/net/imglib2/util/Intervals.java
```diff
@@ -1040,6 +1104,28 @@ public static boolean equalDimensions( final Interval a, final Interval b )
...
+	 * consider using the interval directly. See
+	 * {@link RealInterval#realMin(double[])}.
+	 * </p>
+	 * 
+	 * @param interval
+	 *            something with interval boundaries
+	 * 
+	 * @return minimum as a new RealPoint 
+	 */
+	public static RealPoint minAsRealPoint( final RealInterval interval )
```
axtimwalde, line 1122: may be remove because now we have the defaults

src/main/java/net/imglib2/util/Localizables.java
```diff
@@ -47,12 +47,14 @@
 
 public class Localizables
 {
-
 	public static long[] asLongArray( final Localizable localizable )
```
axtimwalde, line 50: potentially deprecate because we now have the more convenient defaults

src/main/java/net/imglib2/util/Localizables.java
```diff
@@ -47,12 +47,14 @@
...
 	{
-		final long[] result = new long[ localizable.numDimensions() ];
-		localizable.localize( result );
-		return result;
+		return localizable.locationToLongArray();
+	}
+
+	public static double[] asDoubleArray( final RealLocalizable realLocalizable )
```
axtimwalde, line 55: may be deprecate because now we have the more convenient defaults

bogovicj:
Agree with removing / deprecating static methods.

On `positionAs...` vs `locationAs...` 
I went with "location" trying to be consistent with the idea that `localize` / `Localizable` are get-able vs `Positionable`s are set-able, and because that's what you initially suggested. What do you think?

axtimwalde:
Nah, I think the `get[...]` methods are sufficiently leaning `Position` https://javadoc.scijava.org/ImgLib2/net/imglib2/Localizable.html.  `localize(long[])` is the noble exception because it is not a getter and writes the position into an array as side effect.  It's a good word for that, it could also be `position(long[])`, in line with `min` and `max` in `Interval`, but I also feel that somebody had this discussion before and the conclusion was to call it localize...

bogovicj:
Thoughts on deprecating methods in `Intervals` that overlap with what's added here? e.g. [`Intervals.minAsLongArray`](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/util/Intervals.java#L928)

src/main/java/net/imglib2/util/Localizables.java
bogovicj: Because I added this one as part of the PR, I'll remove instead of deprecating.

axtimwalde:
For now, add a to the JavaDoc that there is an alternative now that is more convenient and suggest they should be deprecated.  We will most likely not remove them for a very long time because everybody uses them.

src/main/java/net/imglib2/Interval.java
```diff
@@ -145,48 +145,68 @@ default long dimension( final int d )
...
 	}
 
 	/**
 	 * Allocates a new {@link Point} with the maximum of this Interval.
-	 * 
+	 *
+	 * Please note that his method allocates a new {@link Point} each time
+	 * which introduces notable overhead in both compute and memory.
+	 * If you query it frequently, you should allocate a dedicated array
+	 * first and reuse it with {@link #min(long[])}.
```
bogovicj, line 169: change doc to:
first and reuse it with{@link #min(Positionable)}.

src/main/java/net/imglib2/Interval.java
```diff
@@ -145,48 +145,68 @@ default long dimension( final int d )
...
 	}
 
 	/**
 	 * Allocates a new {@link Point} with the maximum of this Interval.
-	 * 
+	 *
+	 * Please note that his method allocates a new {@link Point} each time
+	 * which introduces notable overhead in both compute and memory.
+	 * If you query it frequently, you should allocate a dedicated array
+	 * first and reuse it with {@link #max(long[])}.
```
bogovicj, line 203: change doc to:
first and reuse it with{@link #max(Positionable)}.

src/main/java/net/imglib2/Localizable.java
```diff
@@ -99,7 +104,12 @@ default int getIntPosition( final int d )
 	/**
 	 * Allocate and return a {@link Point} containing the localizable's
 	 * position.
-	 * 
+	 *
+	 * Please note that his method allocates a new {@link Point} each time
+	 * which introduces notable overhead in both compute and memory.
+	 * If you query it frequently, you should allocate a dedicated array
+	 * first and reuse it with {@link #localize(long[])}.
```
bogovicj, line 111: change doc to:
first and reuse it with{@link #localize(Positionable)}.

src/main/java/net/imglib2/RealInterval.java
```diff
@@ -121,50 +121,69 @@ default void realMax( final RealPositionable max )
...
 
 	/**
-	 * Allocates a new {@link RealPoint} with the maximum of this RealInterval.
-	 * 
+	 * Allocates a new {@link RealPoint} with the minimum of this RealInterval.
+	 *
+	 * Please note that his method allocates a new {@link RealPoint} each time
+	 * which introduces notable overhead in both compute and memory.
+	 * If you query it frequently, you should allocate a dedicated array
+	 * first and reuse it with {@link #realMin(double[])}.
```
bogovicj, line 145: change doc to:
first and reuse it with {@link #realMin(RealPositionable)}.

src/main/java/net/imglib2/RealInterval.java
```diff
@@ -121,50 +121,69 @@ default void realMax( final RealPositionable max )
...
 	}
 
 	/**
 	 * Allocates a new {@link RealPoint} with the maximum of this RealInterval.
-	 * 
+	 *
+	 * Please note that his method allocates a new {@link RealPoint} each time
+	 * which introduces notable overhead in both compute and memory.
+	 * If you query it frequently, you should allocate a dedicated array
+	 * first and reuse it with {@link #realMax(double[])}.
```
bogovicj, line 179: change doc to:
first and reuse it with {@link #realMax(RealPositionable)}.

src/main/java/net/imglib2/Localizable.java
axtimwalde: This method does not exist.

axtimwalde: Added it.

PR #291, opened 2020-07-23
Converters and related classes for IterableRealIntervals

StephanPreibisch:
Support virtual intensity transformations for IterableRealInterval


axtimwalde:
I decided to merge this because waiting for @StephanPreibisch to write the pending tests will delay this forever ;).  It would still be good to add tests, there are good templates for the other accessibles.

PR #292, opened 2020-08-01
Fix RealInterval.maxAsRealPoint

hanslovsky:
The return value `RealPoint max` gets populated via `realMin( max )` in `RealInterval.maxAsRealPoint` when it should be `realMax( max )`.


axtimwalde:
Thanks!

PR #293, opened 2020-08-27
add interpolated stacks of RealRandomAccessibles

axtimwalde:
this enables the composition of independently aligned hyperslices
and later transformation with only one interpolator

A stack of `RealRandomAccessible`s becomes a `RealRandomAccessible`.

While I am very sure about the necessity and usefulness of this, I am unsure if this overstretches the semantic meaning of ImgLib2 interpolators.  I added it here, because I want to support more than one slice interpolator (here it's NN and L, others straight forward to add), however, the source `F` for this interpolator is the rather meaningless `List<RealRandomAccessible>` that is also `EuclideanSpace` so that it fits with what `Interpolant` requires, which is necessary so `Views.interpolate(...)` works.  We could also implement this directly without passing through `Interpolant` which would then be an independent yet somehow related thing which I found even more unsatisfying.


axtimwalde:
May be a bit more detail.  This rewrite of `Interpolant`:
```java
final public class Interpolant< T, F > implements RealRandomAccessible< T >, View
{
	protected final F source;

	protected final int n;

	final InterpolatorFactory< T, F > factory;

	public Interpolant( final EuclideanSpace source, final InterpolatorFactory< T, F > factory )
	{
		this.source = Cast.unchecked( source );
		this.factory = factory;
		this.n = source.numDimensions();
	}

	public Interpolant( final F source, final InterpolatorFactory< T, F > factory, final int n )
	{
		this.source = source;
		this.factory = factory;
		this.n = n;
	}

	@Override
	public int numDimensions()
	{
		return n;
	}
[...]
```
would keep `Views.interpolate(...)` functional but also enable construction with non-`EuclideanSpace` sources which is fine for both the `Interpolator` and `InterpolatorFactory`, however, it gives up compile time type safety because Java does not allow me to narrow down the parameters in the constructor to `<E extends F & EusclideanSpace>`.

axtimwalde:
Questions resolved after in person discussion with @tpietzsch.  It's ok for this to be an interpolator.  I changed `Interpolator` as suggested in the previous comment with deprecated backwards compatible constructor.  Also updated `Views` to use the compile time safe constructor and had to remove the old deprecated `extendValue` method for which less restrictive version under the same name exists, so everything should compile fine against this.

PR #298, opened 2020-11-03
Add IntegerTypes utility class

imagejan:
This commits adds a simple utility class to retrieve an `IntegerType` based on given `min` and `max` values, or on a `max` value and a flag for signed/unsigned.

It includes a test for all possible return types.

Also removed hard-coded version pinning in the test dependencies, and added myself as contributor.

See also [this discussion](https://gitter.im/imglib/imglib2?at=5f97c99ebbffc02b5841671d) on Gitter.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -166,6 +166,11 @@ Jean-Yves Tinevez and Michael Zinsmaier.</license.copyrightOwners>
 			<properties><id>bogovicj</id></properties>
 		</contributor>
 		<contributor>
+			<name>Jan Eglinger</name>
+			<url>https://imagej.net/User:Eglinger</url>
+			<properties><id>imagejan</id></properties>
+		</contributor>
+		<contributor>
 			<name>Jan Funke</name>
 			<url>https://imagej.net/User:Funke</url>
 			<properties><id>funkey</id></properties>
@@ -230,13 +235,11 @@ Jean-Yves Tinevez and Michael Zinsmaier.</license.copyrightOwners>
 		<dependency>
 			<groupId>org.openjdk.jmh</groupId>
 			<artifactId>jmh-core</artifactId>
-			<version>1.19</version>
 			<scope>test</scope>
 		</dependency>
 		<dependency>
 			<groupId>org.openjdk.jmh</groupId>
 			<artifactId>jmh-generator-annprocess</artifactId>
-			<version>1.19</version>
 			<scope>test</scope>
 		</dependency>
 		<dependency>
--- a/src/main/java/net/imglib2/type/numeric/integer/IntegerTypes.java
+++ b/src/main/java/net/imglib2/type/numeric/integer/IntegerTypes.java
@@ -1,0 +1,42 @@
+package net.imglib2.type.numeric.integer;
+
+import net.imglib2.type.logic.BitType;
+import net.imglib2.type.numeric.IntegerType;
+
+public class IntegerTypes
+{
+	private IntegerTypes()
+	{
+		// NB: prevent instantiation of static utility class
+	}
+
+	public static IntegerType< ? > smallestType( boolean signed, long max )
+	{
+		return smallestType( signed ? -1 : 0, max );
+	}
+
+	public static IntegerType< ? > smallestType( long min, long max )
+	{
+		if ( min > max )
+			throw new IllegalArgumentException( "Wrong usage: min (" + min + ") > max (" + max + ")" );
+		if ( min >= 0 )
+		{
+			if ( max <= 1L )
+				return new BitType();
+			if ( max <= 0xffL )
+				return new UnsignedByteType();
+			if ( max <= 0xffffL )
+				return new UnsignedShortType();
+			if ( max <= 0xffffffffL )
+				return new UnsignedIntType();
+			return new UnsignedLongType();
+		}
+		if ( min >= Byte.MIN_VALUE && max <= Byte.MAX_VALUE )
+			return new ByteType();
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -230,13 +235,11 @@ Jean-Yves Tinevez and Michael Zinsmaier.</license.copyrightOwners>
 		<dependency>
 			<groupId>org.openjdk.jmh</groupId>
 			<artifactId>jmh-core</artifactId>
-			<version>1.19</version>
```
mkitti, line 233: Noting that the jmh version can be removed the version is now specified in the scijava pom:
https://github.com/scijava/pom-scijava/commit/2f29fb81acd5b33097ff80b8ab3af4cd33fcbd7f

Merging this would result in jmh being updated to version 1.23
https://github.com/scijava/pom-scijava/blob/7847a06b47b12ae200f70588c4cf573ed3d74e57/pom.xml#L1084

imagejan: Is there any particular reason why `imglib2` still needs to stick to version 1.19? In this case the version override should be done via version `<property>` and not hard-coding. But I think it's fine leaving this to be managed by pom-scijava, no?

tpietzsch: yes, should be managed by pom-scijava 👍

src/main/java/net/imglib2/type/numeric/integer/IntegerTypes.java
```diff
@@ -0,0 +1,42 @@
...
+import net.imglib2.type.numeric.IntegerType;
+
+public class IntegerTypes
+{
+	private IntegerTypes()
+	{
+		// NB: prevent instantiation of static utility class
+	}
+
+	public static IntegerType< ? > smallestType( boolean signed, long max )
```
hanslovsky, line 13: Add JavaDoc to clarify what is meant by `smallest`.

src/main/java/net/imglib2/type/numeric/integer/IntegerTypes.java
```diff
@@ -0,0 +1,42 @@
...
+				return new BitType();
+			if ( max <= 0xffL )
+				return new UnsignedByteType();
+			if ( max <= 0xffffL )
+				return new UnsignedShortType();
+			if ( max <= 0xffffffffL )
+				return new UnsignedIntType();
+			return new UnsignedLongType();
+		}
+		if ( min >= Byte.MIN_VALUE && max <= Byte.MAX_VALUE )
```
hanslovsky, line 34: It'd be nice to use the `getMaxValue` and `getMinValue` methods for the bounds checks because the actual values are an implementation detail that is provided through the interface.

hanslovsky:
What was the rationale to create `IntegerTypes` instead of adding this to `IntegerType`? Is it to follow other ImgLib2 examples, e.g. `Intervals` and `Views`?

imagejan:
@hanslovsky wrote:
> What was the rationale to create `IntegerTypes` instead of adding this to `IntegerType`?

I just followed the pattern of other utility classes in ImgLib2. If others agree, I can add these static methods to the `IntegerType` interface instead.

hanslovsky:
> If others agree, I can add these static methods to the `IntegerType` interface instead.

No need to move from my point of view. Initially, I was going to ask you to move the methods but then I remember that you followed a common pattern in ImgLib2. I don't have any issue with leaving them in `IntegerTypes`.

src/main/java/net/imglib2/type/numeric/integer/IntegerTypes.java
imagejan: Any suggestions how this should be done? In order to check `getMinValue()` and `getMaxValue()`, I first have to create an instance of the respective type, right?
Isn't it the implicit contract of e.g. `ByteType` that it can take values between `Byte.MIN_VALUE` and `Byte.MAX_VALUE`?

hanslovsky: > In order to check `getMinValue()` and `getMaxValue()`, I first have to create an instance of the respective type, right?

Yes, that is correct. I would create a `private static List<? extends IntegerType< ? > >` (one for signed and one for unsigned) sorted by size in ascending order, and then filter for the first element that matches the bounds, or `throw` as you do if there are none, e.g.

```java
signedIntegers
    .stream()
    .filter(i -> v >= i.getMinValue() & v <= i.getMaxValue())
    .findFirst()
    .get() // check and exception would have to come here
    .createVariable()
```

> Isn't it the implicit contract of e.g. `ByteType` that it can take values between `Byte.MIN_VALUE` and `Byte.MAX_VALUE`?

Yes it is but I personally prefer explicitly using the interface methods.

imagejan: Thanks a lot for the suggestion, @hanslovsky.

I implemented it in https://github.com/imglib/imglib2/pull/298/commits/5505beb8c78a059f3c5ac3c66d1b31efb8b45f9a. As we only allow `long` (not e.g. `BigInteger`) for `min` and `max` values, I didn't find a condition in which no type would get selected, so I skipped checking for an empty list and throwing an exception.

src/main/java/net/imglib2/type/numeric/integer/IntegerTypes.java
imagejan: Done. Happy to hear suggestions how to improve/clarify though 🙂 .

hanslovsky: I like your JavaDoc, this is resolved for me

hanslovsky (approved):

PR #299, opened 2020-11-17
Buffer-backed access (full implementation)

### LINKED ISSUE #5: Port NIO-backed images to ImgLib2

mkitti:
# Buffer-backed access (full implementation)

These changes complete #278 and fixes #5 

7cc0468 implements a lightweight interface `DataAccess` with a `createView(Object)` method as per the Twitter discussion. It also includes a default implementation that just returns `this`. The interface definition returns type `Object`, which seems better than my attempt to parameterize generics across the entire `Access` tree.

6f851db contains the `BufferAccess` interface and its implementations. Per @hanslovsky's design in #278 `BufferAccess` implements `ArrayDataAccess` and `VolatileAccess`. The concrete classes share a common `AbstractBufferAccess`.

The `net.imglib2.img.basictypeaccess.nio` tree is as follows:

* BufferAccess implements VolatileAccess, ArrayDataAccess
    * AbstractBufferAccess
        * ByteBufferAccess
        * CharBufferAccess
        * FloatBufferAccess
        * DoubleBufferAccess
        * IntBufferAccess
        * ShortBufferAccess
        * LongBufferAccess

### General
1. `ArrayDataAccess` is implemented as follows
    * `createArray( int numEntities )` allocates a new `java.nio.Buffer` based on whether the current one `isDirect` or not and if the current `Access` `isValid`
    * `getCurrentStorageArray()` returns the raw `Buffer` without duplicating it
    * `getArrayLength()` is returns `Buffer.limit()` (as opposed to `Buffer.capacity()` as in #278)
2. `VolatileAccess` is implemented with support from `AbstractBufferAccess`
3. All concrete classes store a type-specific `Buffer`
    * Per https://github.com/imglib/imglib2/pull/278#issuecomment-565815728 , all concrete classes can also be constructed from a `java.nio.ByteBuffer` and contain a `fromByteBuffer` static method.

### Specific
1. As in https://github.com/imglib/imglib2/pull/278#issuecomment-565822116 and item 1 above, `createArray( int numEntities )` allocates a `Buffer` that has the same directness as the current `Buffer`

## Differences to imglib[1] NIO*Array

The current implementation differs from the previous NIO*Array (e.g. [`NIOLongArray`](https://github.com/fiji/legacy-imglib1/blob/master/src/main/java/mpicbg/imglib/container/basictypecontainer/array/NIOLongArray.java) :
1. The constructor receives a `java.nio.Buffer` directly rather than array. The user can wrap (e.g. `LongBuffer.wrap`) an array if needed.
2. `getCurrentStorageArray` returns a `BufferAccess` rather than an `Array`

## New Items

1. I named the classes `LongBufferAccess` rather than `NIOLongArray` or `LongBufferLongAccess`. The buffers are distinct from arrays. Also the `Buffer` type and `Access` type are always the same in this implementation.
2. I proxied in the bulk relative `get` and `put` methods of `java.nio.LongBuffer` as `getValues` and `setValues` that allows the user to quickly copy in or out data using the corresponding `Array` access (e.g. `AbstractLongArray` as of f5609a6). This is mainly a convenience and could be dropped.
3. `Buffer.duplicate()` is only used in `DataAccess.createView(Object)` and in the bulk `getValues` and `setValues` methods. `Buffer.duplicate()` only actually exists in Java 9+ but is implemented in the concrete classes in Java 8 (e.g. `LongBuffer`)

## Change Log

* f5609a6 Changed to using the `Abstract*Array` types rather than `*Array`.

### DIFF
--- a/src/main/java/net/imglib2/img/array/ArrayImg.java
+++ b/src/main/java/net/imglib2/img/array/ArrayImg.java
@@ -84,7 +84,7 @@ public class ArrayImg< T extends NativeType< T >, A extends DataAccess > extends
 	@Override
 	public A update( final Object o )
 	{
-		return (A) data.createView(o);
+		return ( A ) data.createView( o );
 	}
 
 	@Override
--- a/src/main/java/net/imglib2/img/array/ArrayImg.java
+++ b/src/main/java/net/imglib2/img/array/ArrayImg.java
@@ -39,6 +39,7 @@ import net.imglib2.FlatIterationOrder;
 import net.imglib2.Interval;
 import net.imglib2.img.AbstractNativeImg;
 import net.imglib2.img.Img;
+import net.imglib2.img.basictypeaccess.DataAccess;
 import net.imglib2.type.NativeType;
 import net.imglib2.util.Fraction;
 import net.imglib2.util.IntervalIndexer;
@@ -58,7 +59,7 @@ import net.imglib2.view.iteration.SubIntervalIterable;
  * @author Stephan Preibisch
  * @author Stephan Saalfeld
  */
-public class ArrayImg< T extends NativeType< T >, A > extends AbstractNativeImg< T, A > implements SubIntervalIterable< T >
+public class ArrayImg< T extends NativeType< T >, A extends DataAccess > extends AbstractNativeImg< T, A > implements SubIntervalIterable< T >
 {
 	final int[] steps, dim;
 
@@ -79,10 +80,11 @@ public class ArrayImg< T extends NativeType< T >, A > extends AbstractNativeImg<
 		this.data = data;
 	}
 
+	@SuppressWarnings("unchecked")
 	@Override
 	public A update( final Object o )
 	{
-		return data;
+		return (A) data.createView(o);
 	}
 
 	@Override
--- a/src/main/java/net/imglib2/img/array/ArrayImg.java
+++ b/src/main/java/net/imglib2/img/array/ArrayImg.java
@@ -80,7 +80,7 @@ public class ArrayImg< T extends NativeType< T >, A extends DataAccess > extends
 		this.data = data;
 	}
 
-	@SuppressWarnings("unchecked")
+	@SuppressWarnings( "unchecked" )
 	@Override
 	public A update( final Object o )
 	{
@@ -232,5 +232,21 @@ public class ArrayImg< T extends NativeType< T >, A extends DataAccess > extends
 	public Object subIntervalIterationOrder( final Interval interval )
 	{
 		return new FlatIterationOrder( interval );
+	}
+
+	/**
+	 * Deprecated constructor for binary compatibility when A was not bounded by
+	 * DataAccess
+	 * 
+	 * @param data
+	 *            - will be cast to DataAccess type A
+	 * @param dim
+	 * @param entitiesPerPixel
+	 */
+	@SuppressWarnings( "unchecked" )
+	@Deprecated
+	public ArrayImg( final Object data, final long[] dim, final Fraction entitiesPerPixel )
+	{
+		this( ( A ) data, dim, entitiesPerPixel );
 	}
 }
--- a/src/main/java/net/imglib2/img/basictypeaccess/nio/AbstractBufferAccess.java
+++ b/src/main/java/net/imglib2/img/basictypeaccess/nio/AbstractBufferAccess.java
@@ -61,14 +61,14 @@ public abstract class AbstractBufferAccess< A extends AbstractBufferAccess< A, B
 	/**
 	 * Default valid setting if not specified
 	 */
-	protected final static boolean DEFAULT_IS_VALID = true;
+	static final boolean DEFAULT_IS_VALID = true;
 
 	/**
 	 * If valid or not as per {@link VolatileAccess}.
 	 */
-	protected final boolean isValid;
+	private final boolean isValid;
 
-	protected final B buffer;
+	final B buffer;
 
 	/*
 	 * Constructors
@@ -169,7 +169,7 @@ public abstract class AbstractBufferAccess< A extends AbstractBufferAccess< A, B
 	 * @param isDirect
 	 * @return
 	 */
-	protected ByteBuffer allocateByteBuffer( final int numEntities, final boolean isDirect )
+	ByteBuffer allocateByteBuffer( final int numEntities, final boolean isDirect )
 	{
 		if ( isDirect )
 			return ByteBuffer.allocateDirect( numEntities * getNumBytesPerEntity() );
@@ -185,7 +185,7 @@ public abstract class AbstractBufferAccess< A extends AbstractBufferAccess< A, B
 	 * @param isValid
 	 * @return
 	 */
-	protected A allocate( final int numEntities, final boolean isDirect, final boolean isValid )
+	A allocate( final int numEntities, final boolean isDirect, final boolean isValid )
 	{
 		return newInstance( allocateByteBuffer( numEntities, isDirect ), isValid );
 	}
@@ -197,7 +197,7 @@ public abstract class AbstractBufferAccess< A extends AbstractBufferAccess< A, B
 	 * @param numEntities
 	 * @return
 	 */
-	protected A allocate( final int numEntities )
+	A allocate( final int numEntities )
 	{
 		return allocate( numEntities, isDirect(), isValid() );
 	}
@@ -231,6 +231,6 @@ public abstract class AbstractBufferAccess< A extends AbstractBufferAccess< A, B
 	 *
 	 * @return
 	 */
-	protected abstract B duplicateBuffer( B buffer );
+	abstract B duplicateBuffer( B buffer );
 
 }
--- a/src/main/java/net/imglib2/img/basictypeaccess/nio/AbstractBufferAccess.java
+++ b/src/main/java/net/imglib2/img/basictypeaccess/nio/AbstractBufferAccess.java
@@ -1,0 +1,223 @@
+/*
+ * #%L
+ * ImgLib2: a general-purpose, multidimensional image processing library.
+ * %%
+ * Copyright (C) 2009 - 2020 Tobias Pietzsch, Stephan Preibisch, Stephan Saalfeld,
+ * John Bogovic, Albert Cardona, Barry DeZonia, Christian Dietz, Jan Funke,
+ * Aivar Grislis, Jonathan Hale, Grant Harris, Stefan Helfrich, Mark Hiner,
+ * Martin Horn, Steffen Jaensch, Lee Kamentsky, Larry Lindsey, Melissa Linkert,
+ * Mark Longair, Brian Northan, Nick Perry, Curtis Rueden, Johannes Schindelin,
+ * Jean-Yves Tinevez and Michael Zinsmaier.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ * #L%
+ */
+package net.imglib2.img.basictypeaccess.nio;
+
+import java.nio.Buffer;
+import java.nio.ByteBuffer;
+import java.util.function.Function;
+
+import net.imglib2.img.basictypeaccess.DataAccess;
+import net.imglib2.img.basictypeaccess.volatiles.VolatileAccess;
+
+/**
+ * Base abstract class for {@link DataAccess} implementations that wrap
+ * {@link Buffer}.
+ * 
+ * @author Mark Kittisopikul
+ * @author Philipp Hanslovsky (initial specification)
+ *
+ * @param <A> Subclass used as return type for 
+ * {@link #newInstance(Buffer, boolean)} and
+ * {@link #newInstance(ByteBuffer, boolean)}
+ * 
+ * @param <B> {@link Buffer} subclass
+ */
+public abstract class AbstractBufferAccess <A extends AbstractBufferAccess< A, B >, B extends Buffer> implements BufferAccess< A > {
+	
+	/**
+	 * Default valid setting if not specified
+	 */
+	protected static boolean DEFAULT_IS_VALID = true;
+	
+	/**
+	 * If valid or not as per {@link VolatileAccess}.
+	 */
+	protected final boolean isValid;
+
+	protected final B buffer;
+	
+	/*
+	 * Constructors
+	 */
+
+	public AbstractBufferAccess(final B buffer, final boolean isValid) {
+		this.buffer = buffer;
+		this.isValid = isValid;
+	}
+	
+	public AbstractBufferAccess(final B buffer) {
+		this(buffer, DEFAULT_IS_VALID);
+	}
+	
+	public AbstractBufferAccess(final Function<Integer,B> allocate, final int numEntities) {
+		this( allocate.apply(numEntities) );
+	}
+	
+	/*
+	 * Public methods
+	 */
+	
+	/**
+	 * Returns a new instance of BufferAccess with duplicated Buffer for
+	 * thread safety.
+	 */
+	@Override
+	public A createView(Object o) {
+		return newInstance( duplicateBuffer(buffer), isValid() );
+	}
+	
+	@Override
+	public boolean isValid()
+	{
+		return isValid;
+	}
+	
+	/**
+	 * Respects the directness of the buffer.
+	 * If this buffer is direct, then the new backing buffer is also direct.
+	 */
+	@Override
+	public A createArray( int numEntities )
+	{
+		return allocate( numEntities );
+	}
+
+	/**
+	 * Returns the raw buffer (not duplicated)
+	 */
+	@Override
+	public B getCurrentStorageArray()
+	{
+		return buffer;
+	}
+
+	/**
+	 * Returns the Buffer's limit or zero if buffer is null.
+	 * 
+	 * @see Buffer#limit()
+	 */
+	@Override
+	public int getArrayLength()
+	{
+		if( buffer == null )
+			return 0;
+		else
+			return buffer.limit();
+	}
+	
+	/**
+	 * Return the Buffer's directness or false is the buffer is null.
+	 */
+	@Override
+	public boolean isDirect() {
+		return buffer != null && buffer.isDirect();
+	}
+	
+	@Override
+	public boolean isReadOnly() {
+		return buffer.isReadOnly();
+	}
+	
+	/*
+	 * Protected methods
+	 */
+	
+	/**
+	 * Allocate a new ByteBuffer with initial capacity and directness.
+	 * 
+	 * @param numEntities
+	 * @param isDirect
+	 * @return
+	 */
+	protected ByteBuffer allocateByteBuffer( int numEntities, boolean isDirect ) {
+		if(isDirect)
+			return ByteBuffer.allocateDirect( numEntities * getNumBytes() );
+		else
+			return ByteBuffer.allocate( numEntities * getNumBytes() );
+	}
+	
+	/**
+	 * Allocate a new BufferAccess specifying the directness and validity.
+	 * 
+	 * @param numEntities
+	 * @param isDirect
+	 * @param isValid
+	 * @return
+	 */
+	protected A allocate( int numEntities, boolean isDirect, boolean isValid) {
+		return newInstance( allocateByteBuffer(numEntities, isDirect), isValid );
+	}
+	
+	/**
+	 * Allocate a new BufferAccess using the current buffer's directness
+	 * and validity.
+	 * 
+	 * @param numEntities
+	 * @return
+	 */
+	protected A allocate( int numEntities ) {
+		return allocate( numEntities, isDirect(), isValid() );
+	}
+	
+	/*
+	 * Abstract methods
+	 */
+	
+	/**
+	 * Get number of bytes for a specific primitive type.
+	 * 
+	 * This usually retrieves a static field.
+	 */
+	public abstract int getNumBytes();
+	
+	/**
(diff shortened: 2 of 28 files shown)


mkitti:
I went a bit further to parameterize `DataAccess` and other subinterfaces, but this seemed too much and is not included here.
https://github.com/mkitti/imglib2/commit/5bd41a20fbede445017e773fda47eeb2be72a64e

hanslovsky:
Beautiful PR! I will have a look, baby boy permitting

mkitti:
Thanks @hanslovsky . I really appreciated your initial notes on this issue.

src/main/java/net/imglib2/img/array/ArrayImg.java
```diff
@@ -79,10 +80,11 @@ public ArrayImg( final A data, final long[] dim, final Fraction entitiesPerPixel
 		this.data = data;
 	}
 
+	@SuppressWarnings("unchecked")
 	@Override
 	public A update( final Object o )
 	{
-		return data;
+		return (A) data.createView(o);
```
hanslovsky, line 87: Does not conform with imglib2 formatting. I saw that in a few other places. Probably best to run the formatter on all files.

src/main/java/net/imglib2/img/basictypeaccess/nio/AbstractBufferAccess.java
```diff
@@ -0,0 +1,223 @@
...
+ * {@link #newInstance(ByteBuffer, boolean)}
+ * 
+ * @param <B> {@link Buffer} subclass
+ */
+public abstract class AbstractBufferAccess <A extends AbstractBufferAccess< A, B >, B extends Buffer> implements BufferAccess< A > {
+	
+	/**
+	 * Default valid setting if not specified
+	 */
+	protected static boolean DEFAULT_IS_VALID = true;
```
hanslovsky, line 61: Should be `final`.

src/main/java/net/imglib2/img/basictypeaccess/nio/BufferAccess.java
```diff
@@ -0,0 +1,87 @@
...
+	 * @see Buffer#isReadOnly()
+	 */
+	public boolean isReadOnly();
+	
+	/**
+	 * Number of bytes for this type
+	 * 
+	 * @return
+	 */
+	public int getNumBytes();
```
hanslovsky, line 76: This may be confused for the total number of bytes in the buffer. Consider changing the name to be more specific, e.g. `getNumBytesPerEntity` or be specific in the doc.

src/main/java/net/imglib2/img/basictypeaccess/nio/AbstractBufferAccess.java
```diff
@@ -0,0 +1,223 @@
...
+	/**
+	 * Allocate a new BufferAccess specifying the directness and validity.
+	 * 
+	 * @param numEntities
+	 * @param isDirect
+	 * @param isValid
+	 * @return
+	 */
+	protected A allocate( int numEntities, boolean isDirect, boolean isValid) {
+		return newInstance( allocateByteBuffer(numEntities, isDirect), isValid );
```
hanslovsky, line 179: Hard-coding this to `ByteBuffer` might be an issue:
 1. `newInstance` is generic wrt to the buffer. What happens if we pass a `ByteBuffer` but the generic bound `B` is `DoubleBuffer`?
 2. For any type that has multiple bytes per entity, `numEntities` is restricted to about `Integer.MAX_VALUE / getNumBytes()`. For `double`, that would mean only 1/8 of the capacity compared to `byte`.

hanslovsky:
I left a few comments. Overall, this looks good to me. My only real concern is that the buffer type is hard-coded to `ByteBuffer` in `AbstractBufferAccess.allocate`, cf https://github.com/imglib/imglib2/pull/299#discussion_r526547213

I will leave the more thorough review to the active maintainers.

src/main/java/net/imglib2/img/basictypeaccess/nio/AbstractBufferAccess.java
mkitti: 1. There are currently two `newInstance` methods. One is specified by the interface which is only generic to the `Access` type. Another is specified by `AbstractBufferAccess` which is also generic to the `java.nio.Buffer` implementation.

```java
public A newInstance( ByteBuffer buffer, boolean isValid ); // specified by BufferAccess < A >

public abstract A newInstance(B buffer, boolean isValid ); // specified by AbstractBufferAccess <A , B >
```

The idea is that every `BufferAccess` can either be created by either providing a `ByteBuffer` or a more specific `Buffer` like `LongBuffer`. If given a `ByteBuffer`, then the `ByteBuffer` is converted into the type-specific buffer (e.g. `LongBuffer`)

2. This machinery exist mainly to support `ArrayDataAccess.createArray( int numEntities )`. There is a Java limitation in the language itself in that currently you can only create a _direct_ `Buffer` by first creating a `ByteBuffer` and then transforming it into a more specific `Buffer`. Only `ByteBuffer` has `allocateDirect` as you pointed out earlier ( https://github.com/imglib/imglib2/pull/278#issue-353261538 ).

For a _non-direct_ `Buffer` you could get around this by just constructing a `BufferAccess` by passing a `LongBuffer`.

The solution then is that concrete implementations of `AbstractBufferAccess` should overload `allocate` to create a _non-direct_ of a longer size if available. This is already done by the concrete class constructors. I'll revise this in the concrete classes to make longer buffers available to `ArrayDataAccess.createArray( int numEntities )`, but I think I'll keep this `AbstractBufferAccess` implementation here.

src/main/java/net/imglib2/img/basictypeaccess/nio/BufferAccess.java
mkitti: Good idea. I'll rename the constants as well.

src/main/java/net/imglib2/img/basictypeaccess/nio/AbstractBufferAccess.java
mkitti: Agreed.

src/main/java/net/imglib2/img/array/ArrayImg.java
mkitti: I could use some help with this. How do I access the formatter? Is this just the generic Eclipse formatter or is there a imglib2 format ruleset I'm missing?

src/main/java/net/imglib2/img/basictypeaccess/nio/AbstractBufferAccess.java
hanslovsky: My first point is resolved, I did not see the `newInstance( ByteBuffer, boolean )` signature, should've opened in an IDE instead of just looking at the diff. As to my second point, how about adding an abstract method
```java
protected abstract B allocateBuffer( int numEntities, boolean isDirect );
```
and then change `allocate` to
```java
protected A allocate( int numEntities, boolean isDirect, boolean isValid) {
	return newInstance( allocateBuffer(numEntities, isDirect), isValid );
}
```

In the concrete implementations, e.g. `DoubleBuffer`, implement `allocateBuffer` like this:
```java
@Override
protected DoubleBuffer allocateBuffer(int numEntities, boolean isDirect) {
	if(isDirect)
		return ByteBuffer.allocateDirect( numEntities * getNumBytes() ).asDoubleBuffer();
	else
		return DoubleBuffer.allocate( numEntities );
}
```

Then, at least for non-direct buffers, we can use the maximum `numEntitites`/capacity as expected. For the `isDirect` branch, maybe we should add a check and throw an exception if `numEntitites` is too large.

src/main/java/net/imglib2/img/array/ArrayImg.java
hanslovsky: I typically use the [Eclipse formatter](https://github.com/imglib/imglib2/blob/master/doc/imglib2-eclipse-mars-clean-up-formatter-style.xml). I have not had the best experience with it in IntelliJ but I see there was a [commit to add a formatter for IntelliJ](https://github.com/imglib/imglib2/commit/9998be95f18962c4e8a09f4afdb7502a30f0eded) just three months ago.

src/main/java/net/imglib2/img/basictypeaccess/nio/DoubleBufferAccess.java
```diff
@@ -0,0 +1,238 @@
...
+	 * since ByteBuffer is restricted to Integer.MAX_VALUE entities.
+	 */
+	@Override
+	protected DoubleBufferAccess allocate( final int numEntities, final boolean isDirect, final boolean isValid )
+	{
+		if ( isDirect )
+			return super.allocate( numEntities, isDirect, isValid );
+		else
+			return new DoubleBufferAccess( numEntities, isValid );
+	}
```
mkitti, line 149: @hanslovsky I ended overloading allocate like this before I saw your comment.

src/main/java/net/imglib2/img/basictypeaccess/nio/BufferAccess.java
mkitti: Resolved as of b4f1233

src/main/java/net/imglib2/img/basictypeaccess/nio/AbstractBufferAccess.java
mkitti: Resolved as of b4f1233

src/main/java/net/imglib2/img/array/ArrayImg.java
mkitti: Resolved and formatted as of b4f1233

src/main/java/net/imglib2/img/basictypeaccess/nio/DoubleBufferAccess.java
hanslovsky: That works for me, too.

src/main/java/net/imglib2/img/basictypeaccess/nio/AbstractBufferAccess.java
hanslovsky: Fixed in 83c61ea

hanslovsky (approved):
The proposed changes look good to me but I will leave further review/merging to the active maintainers

tpietzsch:
I pushed some changes to https://github.com/imglib/imglib2/commits/nio_buffers:

- Remove unnecessary restriction to `A extends DataAccess` from `Cell<A>`.
- Changed protected to package-private for classes in `net.imglib2.img.basictypeaccess.nio`. Please reinstate protected if you need to derive additional classes outside imglib2 core.
- additional formatting.

If you agree with these changes please merge into your PR.
Otherwise this looks good to go.

I'm not sure whether the replacement of `A` by `A extends DataAccess` breaks binary compatibility (but I don't think so)?

mkitti:
On binary compatibility, I may have come across some examples where I had very generic code break because A was previously unconstrained and the method signatures no longer matched. This may be an extreme case, however.

tpietzsch:
> On binary compatibility, I may have come across some examples where I had very generic code break because A was previously unconstrained and the method signatures no longer matched. This may be an extreme case, however.

Yes, that is definitely a problem. It happened to us before, e.g. https://github.com/imglib/imglib2/pull/280

Also here, the erasure of `A` changes from `Object` to `DataAccess`.
In this case we might be safe because the `A` only appears as a type parameter in e.g. `ArrayImg<T,A>` and that is erased to `ArrayImg` no matter what. But maybe I'm overlooking something unrelated to the changed classes.

mkitti:
I went back a few commits, and I think the issue I encountered was more due to having multiple versions of imglib2 on my path at the time.

Might there be a binary compatibility issue with the constructor of `ArrayImg`?

```java
	public static < T extends NativeType< T >, A extends DataAccess > ArrayImg< T, A > createArrayImg(
			final NativeTypeFactory< T, A > factory, final A data, final long... dims)
        {
		final Fraction fraction = factory.createLinkedType(null).getEntitiesPerPixel();
		final ArrayImg< T, A > img = new ArrayImg< T, A >(data, dims, fraction);
		final T linkedType = factory.createLinkedType(img);
		img.setLinkedType(linkedType);
		return img;
	}
```

mkitti:
After reading more about binary compatibility, it seems unwise to add a bound to `A`. Rather we should check if `A` is a `DataAccess` and invoke `createView` only then:

```java
public class ArrayImg< T extends NativeType< T >, A > extends AbstractNativeImg< T, A > implements SubIntervalIterable< T >
...
	@SuppressWarnings("unchecked")
	@Override
	public A update( final Object o )
	{
		if(data instanceof DataAccess)
			return ( A ) ( (DataAccess) data).createView( o );
		else
			return data;
	}
```

https://docs.oracle.com/javase/specs/jls/se8/html/jls-13.html#jls-13.4.13

> 13.4.13. Method and Constructor Type Parameters
> Adding or removing a type parameter of a method or constructor does not, in itself, have any implications for binary compatibility.
> 
> If such a type parameter is used in the type of the method or constructor, that may have the normal implications of changing the aforementioned type.
> 
> Renaming a type parameter of a method or constructor has no effect with respect to pre-existing binaries.
> 
> Changing the first bound of a type parameter of a method or constructor may change the erasure (§4.6) of any member that uses that type parameter in its own type, and this may affect binary compatibility. Specifically:
> 
> If the type parameter is used as the type of a field, the effect is as if the field was removed and a field with the same name, whose type is the new erasure of the type variable, was added.
> 
> If the type parameter is used as the type of any formal parameter of a method, but not as the return type, the effect is as if that method were removed, and replaced with a new method that is identical except for the types of the aforementioned formal parameters, which now have the new erasure of the type parameter as their type.
> 
> If the type parameter is used as a return type of a method, but not as the type of any formal parameter of the method, the effect is as if that method were removed, and replaced with a new method that is identical except for the return type, which is now the new erasure of the type parameter.
> 
> If the type parameter is used as a return type of a method and as the type of one or more formal parameters of the method, the effect is as if that method were removed, and replaced with a new method that is identical except for the return type, which is now the new erasure of the type parameter, and except for the types of the aforementioned formal parameters, which now have the new erasure of the type parameter as their types.
> 
> Changing any other bound has no effect on binary compatibility.

mkitti:
I've removed bounds on `A` so there are no longer binary compatibility issues. Type erasure should just make `A` an `Object` in most cases.

tpietzsch:
Hmm, having an `instanceof` check per invocation may be a slight performance problem. `ArrayImg` only does it once per accessor, but for example CellImg does it everytime an accessor moves into a new cell, which can be relatively frequently. I would prefer minor API breakage over this, or benchmarks that show that there is no performance impact.

mkitti:
I moved the `instanceof` checks into the constructor.

In `ArrayImg` I just created an anonymous `DataAccess` to wrap around an arbitrary object if it is not a `DataAccess`. One consequence of this design is that `DataAccess.createView` now just returns an `Object`. If this needs to be constrained at some point, a parameterized sub-interface might make sense in the future.

For `AbstractCellImg`, I use a `BiFunction`. If `A` is a `DataAccess`, the `BiFunction` uses `createView`. Otherwise, the `BiFunction` just returns `data`. I end up using `imgOfCells.firstElement().getData()` in the constructor though to check if `A` is a `DataAccess`, which is not so nice for `LazyCellImg`.

Let me know if you think these measures are worth saving API breakage for or not.

mkitti:
I do not see a major performance regression when running the tests.

```junit5
PS>git branch --show-current
master
PS>git rev-parse HEAD
d54bf15cb17d5a22599c687474c267efa573e472
PS>java -jar '..\..\junit-platform-console-standalone-1.7.0.jar' -cp $env:CLASSPATH -p net.imglib2.img.cell

Thanks for using JUnit! Support its development at https://junit.org/sponsoring

.
+-- JUnit Jupiter [OK]
'-- JUnit Vintage [OK]
  +-- CellContainerFactoryTest [OK]
  | +-- testBitDefaultCellDimensions [OK]
  | +-- testFloatDefaultCellSize [OK]
  | +-- testIntDefaultCellDimensions [OK]
  | +-- testFloatDefaultCellDimensions [OK]
  | +-- testIntDefaultCellSize [OK]
  | '-- testBitDefaultCellSize [OK]
  +-- CellImgTest [OK]
  | +-- testCellImgInvalidDimensions [OK]
  | '-- testCellImg [OK]
  +-- CellTest [OK]
  | +-- testLocalIndexCalculation [OK]
  | '-- testGlobalPositionCalculation [OK]
  +-- CopyTest [OK]
  | +-- testCopyToArrayContainerWithDestIteration [OK]
  | +-- testCopyToCellContainerWithDestIteration [OK]
  | +-- testCopyArrayToArrayWithIterationBoth [OK]
  | +-- testCopyToCellContainerWithSourceIteration [OK]
  | '-- testCopyToArrayContainerWithSourceIteration [OK]
  +-- CellContainerTest [OK]
  | '-- equalIterationOrder [OK]
  '-- CellCursorTest [OK]
    +-- testResetWithCursor [OK]
    +-- testSumWithLocalizingCursor [OK]
    +-- testResetWithLocalizingCursor [OK]
    +-- testSumWithRandomAccess [OK]
    +-- testJmpWithLocalizingCursor [OK]
    +-- testSumWithCursor [OK]
    '-- testJmpWithCursor [OK]

Test run finished after 2171 ms
[         8 containers found      ]
[         0 containers skipped    ]
[         8 containers started    ]
[         0 containers aborted    ]
[         8 containers successful ]
[         0 containers failed     ]
[        23 tests found           ]
[         0 tests skipped         ]
[        23 tests started         ]
[         0 tests aborted         ]
[        23 tests successful      ]
[         0 tests failed          ]

PS>git checkout nio_buffers
Switched to branch 'nio_buffers'
PS>git rev-parse HEAD
996fdec583000669d5a8d223c560bcd45ae43ccf

[... rebuild ...]

PS>java -jar '..\..\junit-platform-console-standalone-1.7.0.jar' -cp $env:CLASSPATH -p net.imglib2.img.cell

Thanks for using JUnit! Support its development at https://junit.org/sponsoring

.
+-- JUnit Jupiter [OK]
'-- JUnit Vintage [OK]
  +-- CellContainerFactoryTest [OK]
  | +-- testBitDefaultCellDimensions [OK]
  | +-- testFloatDefaultCellSize [OK]
  | +-- testIntDefaultCellDimensions [OK]
  | +-- testFloatDefaultCellDimensions [OK]
  | +-- testIntDefaultCellSize [OK]
  | '-- testBitDefaultCellSize [OK]
  +-- CellImgTest [OK]
  | +-- testCellImgInvalidDimensions [OK]
  | '-- testCellImg [OK]
  +-- CellTest [OK]
  | +-- testLocalIndexCalculation [OK]
  | '-- testGlobalPositionCalculation [OK]
  +-- CopyTest [OK]
  | +-- testCopyToArrayContainerWithDestIteration [OK]
  | +-- testCopyToCellContainerWithDestIteration [OK]
  | +-- testCopyArrayToArrayWithIterationBoth [OK]
  | +-- testCopyToCellContainerWithSourceIteration [OK]
  | '-- testCopyToArrayContainerWithSourceIteration [OK]
  +-- CellContainerTest [OK]
  | '-- equalIterationOrder [OK]
  '-- CellCursorTest [OK]
    +-- testResetWithCursor [OK]
    +-- testSumWithLocalizingCursor [OK]
    +-- testResetWithLocalizingCursor [OK]
    +-- testSumWithRandomAccess [OK]
    +-- testJmpWithLocalizingCursor [OK]
    +-- testSumWithCursor [OK]
    '-- testJmpWithCursor [OK]

Test run finished after 2177 ms
[         8 containers found      ]
[         0 containers skipped    ]
[         8 containers started    ]
[         0 containers aborted    ]
[         8 containers successful ]
[         0 containers failed     ]
[        23 tests found           ]
[         0 tests skipped         ]
[        23 tests started         ]
[         0 tests aborted         ]
[        23 tests successful      ]
[         0 tests failed          ]
```

tpietzsch:
> Let me know if you think these measures are worth saving API breakage for or not.

Without these latest changes the code is more understandable. I vote for API breakage. I think the fallout downstream will be minor if at all.

@axtimwalde @StephanPreibisch what do you think?

mkitti:
To summarize, 
1. When using `NativeImg.update` we need to return a view of the underlying `java.nio.Buffer` since the `Buffer` classes contain state which is problematic for thread safety.
2. We implemented a `DataAccess` interface with a default method `createView` that just returns `this`.
3. Implementations of `NativeImg.update` would now return the result of `DataAccess.createView` 
4. Most of the access classes in imglib2 now implement `DataAccess`.
5. The simple implementation is add a bound to the type parameter `A` of `ArrayImg` and `AbstractCellImg` so that `A extends DataAccess`.
6. The issue is that this changes the method signatures since type erasure of `A` results in `DataAccess` rather than `Object` potentially breaking binary compatibility.
7. The resulting code, however, is simple.
https://github.com/imglib/imglib2/compare/master...mkitti:nio_buffers_simple

mkitti:
If we do not want to break binary compatibility, then the solution is
1. Do not bound `A` in `ArrayImg` and `AbstractCellImg` so method and constructor signatures do not change.
2. Check if `A instanceof DataAccess` in the `ArrayImg` or `AbstractArrayImg` constructor.
3. Provide distinct implementations of `NativeImg.update` depending on whether `A instanceof DataAccess` or not.

axtimwalde:
I prefer breaking API over `instanceof`.

mkitti:
OK, I forced pushed back to c6a9d07 by @tpietzsch .

Might alternate constructors like the following help binary compatibility?
```
	@SuppressWarnings( "unchecked" )
        @Deprecated
	public ArrayImg( final Object data, final long[] dim, final Fraction entitiesPerPixel ) {
		this( ( A ) data, dim, entitiesPerPixel );
	}
```

```
	@SuppressWarnings( "unchecked" )
	@Deprecated
	public ArrayImgAWTScreenImage( final T type, final Object data, final long[] dim ) {
		this( type, (A) data, dim );
	}
```

Let me see if I can write a test case.

tpietzsch:
> Might alternate constructors like the following help binary compatibility?

that's a great idea

mkitti:
After adding the additional constructors, there are no longer any binary compatibility issues. For the `update` method, Java generates two signatures, one of which still returns `Object`.

I used the following code to test binary compatibility:
```java
		ShortAccess data = new ShortArray(new short[] { 1, 2, 3, 4, 5});
		
		long[] dim = new long[] { 5 };
		ArrayImg< UnsignedShortType, ShortAccess > img = new ArrayImg<>(data, dim, new Fraction());
		UnsignedShortType type = new UnsignedShortType( img );
		img.setLinkedType(type);
		
		ShortAccess updatedData = img.update(this);
		System.out.println(updatedData);
```

I put this in the Sandbox class from the imglib2-tutorials. Sandbox should be compiled against imglib2 via the scijava parent pom. I then I ran it directly from the command line with distinct exported jars of imglib2 from Eclipse.

Without adding the constructors taking `Object` as a parameter in lieu of `A`, I get the following error:
```
Exception in thread "main" java.lang.NoSuchMethodError: net.imglib2.img.array.ArrayImg.<init>(Ljava/lang/Object;[JLnet/imglib2/util/Fraction;)V
```

Adding the constructors, everything seems to work fine. I get the following as output.
```
net.imglib2.img.basictypeaccess.array.ShortArray@668bc3d5
```

I expected to encounter issues with the update method, so I used `javap -c Sandbox.class` to see the bytecode:
```
      75: invokevirtual #30                 // Method net/imglib2/img/array/ArrayImg.setLinkedType:(Lnet/imglib2/type/NativeType;)V
      78: aload_3
      79: aload_0
      80: invokevirtual #34                 // Method net/imglib2/img/array/ArrayImg.update:(Ljava/lang/Object;)Ljava/lang/Object;
      83: checkcast     #38                 // class net/imglib2/img/basictypeaccess/ShortAccess
```

`Sandbox` is calling `update` with `Object` parameter and an `Object` return type.

I then ran `javap -s ArrayImg.class` and discovered the following:
```java
Compiled from "ArrayImg.java"
public class net.imglib2.img.array.ArrayImg<T extends net.imglib2.type.NativeType<T>, A extends net.imglib2.img.basictypeaccess.DataAccess> extends net.imglib2.img.AbstractNativeImg<T, A> implements net.imglib2.view.iteration.SubIntervalIterable<T> {
...
  public A update(java.lang.Object);
    descriptor: (Ljava/lang/Object;)Lnet/imglib2/img/basictypeaccess/DataAccess;
...
  public java.lang.Object update(java.lang.Object);
    descriptor: (Ljava/lang/Object;)Ljava/lang/Object;
}
```

ArrayImg contains two distinct `update` method signatures differing only in return type! This is because `ArrayImg` implements the interface `NativeImg` which after type erasure specifies an `update` method that returns `Object`. [Java implements co-variant return types by including both method signatures](https://blogs.oracle.com/sundararajan/covariant-return-types-in-java).

My conclusion is that with the added constructors in cacce9f, there are no longer binary compatibility issues with this pull request.

hanslovsky:
@tpietzsch Do you think this can be merged?

axtimwalde:
We should merge this and fix problems as they emerge.

Issue #300, opened 2021-01-22
Empty Intervals - numElements and intersection


bogovicj:
I'd like to detect when the intersection of two intervals is empty, and I came across some surprising behavior.

```
// make an interval
FinalInterval A = new FinalInterval( 
		new long[]{ 0, 0, 0 },
		new long[]{ 5, 5, 5 });

// make B such that A intersect B is empty
FinalInterval B = new FinalInterval( 
		new long[]{ 10, 0, 0 }, 
		new long[] { 15, 5, 5 });

// compute the intersection
FinalInterval AandB = Intervals.intersect( A, B );
System.out.println( AandB );

// how many elements in the intersection
System.out.println( Intervals.numElements( AandB ));
```

The output is:
```
FinalInterval [(10, 0, 0) -- (5, 5, 5) = -4x6x6]
-144
```

I was surprised to see `Intervals.numElements` return a negative number.

If the convention is "anything with zero or fewer elements is empty" that's fine, but this output is still surprising, which is why I'd prefer `numElements` return always a non-negative number.  

As well, could we add to the `Interval` documentation something like: 
> An `Interval` for which `max(d)` < `min(d)` for any d, should be treated as empty.

bogovicj:
> "anything with zero or fewer elements is empty"

This does not work with the current intersection implementation because:
```
FinalInterval A = new FinalInterval( 
		new long[]{ 0, 0, 0 },
		new long[]{ 5, 5, 5 });

// make B such that A intersect B is empty
FinalInterval B = new FinalInterval( 
		new long[]{ 10, 0, 10 }, 
		new long[] { 5, 5, 5 });

// compute the intersection
FinalInterval AandB = Intervals.intersect( A, B );
System.out.println( Intervals.numElements( AandB ));
```
returns `96`

i.e., the number of elements of the intersection of two non-intersecting intervals currently can return a positive number.

imagejan:
I think the misconception here is that you call the constructor of `B` with a `min` that is larger than `max` in (at least) one dimension.

The following (Groovy) script works as expected for me:

```groovy
import net.imglib2.FinalInterval
import net.imglib2.util.Intervals

// make an interval
FinalInterval A = new FinalInterval( 
		new long[]{ 0, 0, 0 },
		new long[]{ 5, 5, 5 });

// make B such that A intersect B is empty
FinalInterval B = new FinalInterval( 
		new long[]{ 6, 0, 0 }, 
		new long[] { 10, 5, 5 });

[...]

PR #301, opened 2021-01-22
numElements returns zero for empty Intervals

bogovicj:
Changes the implementation of `Intervals.numElements` so that it returns zero "empty" `Interval`s 
Where an `Interval` is "empty" if its `max(d)` < `min(d)` for any `d`

If we agree, then this definition should be documented.

see #300

### DIFF
--- a/src/main/java/net/imglib2/util/Intervals.java
+++ b/src/main/java/net/imglib2/util/Intervals.java
@@ -545,13 +545,16 @@ public class Intervals
 	 *
 	 * Create a {@link FinalInterval} that represents that interval.
 	 *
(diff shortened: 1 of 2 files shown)


hanslovsky (approved):

tpietzsch:
Looks good to me.
I would maybe change the `a>0 : a ? 0` constructs to `Math.max(a,0)`.

bogovicj:
@tpietzsch 

The latest push adds changes to the behavior of `Intervals.union`, fixing its behavior for empty intervals.
I called the methods with the old behavior `unionUnsafe`.

Alternatively, we could call the new methods `unionSafe`, and leave the old methods untouched and un-renamed.

PR #305, opened 2021-03-22
WIP Add benchmarks to measure performance problems for View.translate() etc.

maarzt:
The problems with the just in time compiler code optimization can
be potentially be triggered for:
* Converts.convert()
* Views.translate()
* Views.extendBorder() etc.

### DIFF
--- a/src/test/java/net/imglib2/converter/ConvertersBenchmark.java
+++ b/src/test/java/net/imglib2/converter/ConvertersBenchmark.java
@@ -1,0 +1,139 @@
+package net.imglib2.converter;
+
+import net.imglib2.RandomAccess;
+import net.imglib2.RandomAccessibleInterval;
+import net.imglib2.img.array.ArrayImgs;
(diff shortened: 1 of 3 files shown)


maarzt:
@tpietzsch Here are that benchmarks that I talked about.

(The PR is intended to be a draft PR but I clicked the wrong button :upside_down_face:)

tpietzsch:
Great! I think, it's very important to have these benchmarks to track progress.
Do you think any of these problems could/should be solved with `ClassCopyProvider`?

tpietzsch:
Can you please format it with imglib2 code style, then I would just merge it

maarzt:
I formatted the files according to the imglib2 coding style. So it can be merged now.

`ClassCopyProvider` could potentially be used to work around this performance problems. But that's not going to be easy, as this might cause new performance problems.

axtimwalde:
Can this be merged?

PR #312, opened 2022-02-25
Add RandomAccessibleInterval.getType

hanslovsky:
Adding `getType` to the `RandomAccessibleInterval` interface would allow individual implementations to provide more efficient implementations. For example, loading the voxel at the min of the RAI may be expensive for some implementations of `NativeImg`. Instead, those implementations can use their `linkedType` to create a copy of the associated type, if available.

[Discussion on gitter for reference](https://gitter.im/imglib/imglib2?at=62100622d1b64840db2a7ca3)

Open questions:
 - We cannot deprecate Utils.getTypeFromInterval because it does not use RAI as argument. We could change the argument to RAI but this would be a breaking change (see comment in JavaDoc for that function)
 - What about getType in other places like RA, RRA, RRARI?
 - Name preference getType vs type?


gselzer:
> * We cannot deprecate Utils.getTypeFromInterval because it does not use RAI as argument. We could change the argument to RAI but this would be a breaking change (see comment in JavaDoc for that function)

Is there a reason not to move this method into `RandomAccessible`? You could still have the default implementation in `RAI`, and then you could deprecate `getTypeFromInterval`.

> * What about getType in other places like RA, RRA, RRARI?

If you moved this method to RA, then you'd get it in all of these places.

> * Name preference getType vs type?

I like `type`, it is shorter and aligns with other methods you might call on a `RAI` (like `min`, `max`, `dimension`, `randomAccess`, etc.)

gselzer:
> Is there a reason not to move this method into `RandomAccessible`?

Ah, you'd want a default implementation in `RandomAccessible` if you added it, that is a good reason. You could hack it with `Views.interval(this, <arbitrary interval>).type()` :grinning:

hanslovsky:
Two additional commits:
 - [7be3635](https://github.com/imglib/imglib2/pull/312/commits/7be3635abc9d7ee9eac7095cf05da624f848d4f0) adds `getType()` to interfaces RA, RRA, RRARI
 - [1f67b63](https://github.com/imglib/imglib2/pull/312/commits/1f67b6375f5a398740f44d9131d2ea0ce897e402) adds more efficient implementation of `getType()` to some classes

hanslovsky:
@gselzer 

> Ah, you'd want a default implementation in` RandomAccessible` if you added it, that is a good reason. You could hack it with `Views.interval(this, <arbitrary interval>).getType()`

Yes, that is correct. You can see how I solved this in [7be3635](https://github.com/imglib/imglib2/pull/312/commits/7be3635abc9d7ee9eac7095cf05da624f848d4f0)

gselzer:
> @gselzer
> 
> > Ah, you'd want a default implementation in` RandomAccessible` if you added it, that is a good reason. You could hack it with `Views.interval(this, <arbitrary interval>).getType()`
> 
> Yes, that is correct. You can see how I solved this in [7be3635](https://github.com/imglib/imglib2/pull/312/commits/7be3635abc9d7ee9eac7095cf05da624f848d4f0)

Aha, `randomAccess().getType()` is a better solution. Looks great @hanslovsky!

hanslovsky:
Adding more implementations for `getType()`, I realized that using a `default` implementation may not be very useful. This already breaks if a wrapping `RAI` does not override it, e.g. something like this is possibly slow if the converted `RAI` does not implement a good way to provide `getType()`:
```java
RAI<T> raiWithEfficientGetType = ...
raiWithEfficientGetType.getType(); // this is fast
RAI<U> notEfficient = Converters.convert(raiWithEfficientGetType, ...);
notEfficient.getType(); // this is slow
```

Several options:
 1. make it not `default`. Probably a big no. All downstream libraries will have to implement this, unless it is covered by some abstract classes in imglib2 core
 2. implement it for all concrete and abstract classes in imglib2 core. This should cover most of what people use regularly.
 3. discard the changes. This would be sad, too.
Might need to go with option (2).

hanslovsky:
In dc0b039, I added an example of how an implementation for (most) of RA/RAI could look like. This just for demonstration purposes and can be easily reverted. I did not override `getType` for these classes:
```
net.imglib2.img.cell.LazyCellImg.LazyCells
net.imglib2.img.cell.CellGrid.CellIntervals
net.imglib2.img.list.ListImg
net.imglib2.util.Localizables.LocationRandomAccessible
net.imglib2.view.FunctionView
net.imglib2.view.composite.CompositeView
net.imglib2.view.composite.CompositeIntervalView
net.imglib2.converter.readwrite.WriteConvertedIterableRandomAccessibleInterval
net.imglib2.view.HyperSlicesView
net.imglib2.converter.read.ConvertedRandomAccessible
net.imglib2.img.cell.LazyCellImg.LazyCells
net.imglib2.img.cell.CellGrid.CellIntervals
net.imglib2.img.list.ListImg
net.imglib2.util.Localizables.LocationRandomAccessible
net.imglib2.view.FunctionView
net.imglib2.view.composite.CompositeView
net.imglib2.view.composite.CompositeIntervalView
net.imglib2.converter.readwrite.WriteConvertedIterableRandomAccessibleInterval
net.imglib2.view.HyperSlicesView
net.imglib2.converter.read.ConvertedRandomAccessible
```

gselzer:
> Adding more implementations for `getType()`, I realized that using a `default` implementation may not be very useful. This already breaks if a wrapping `RAI` does not override it, e.g. something like this is possibly slow if the converted `RAI` does not implement a good way to provide `getType()`:

What do you mean here, @hanslovsky? How does it break?

hanslovsky:
@gselzer Let me provide a different example: A use case that probably is not that uncommon is to take an smaller interval from a larger `RAI`.
Let's assume that the large `RAI` is a cached cell img that is backed by a slow loader, e.g. fetching data over the internet.
```java
RAI<T> cachedImg = ...
```
the cached cell img happens to override `getType()` efficiently in this example, so
```java
cachedImg.getType()
```
is a very fast operation. The user now goes ahead and takes a smaller interval out of this cached cell img, i.e.
```
RAI<T> interval = Views.interval(cachedImg, ...);
```
The user also knows that `cachedImg.getType()` is very efficient and expects the same for `interval`. In this example, the class returned by `Views.interval` does not override `getType()` and falls back to the default implementaiton, i.e. go through the random access and actually load the data from the cached cell img that backs this smaller interval. Hence,
```java
interval.getType()
```
is very slow, and the user has no way of making it fast, again.

This is, of course a hypothetical example, but might easily happen if we provide a default implementation for `getType()`.

gselzer:
Huh, in the case of `View`s, can they not just delegate to the `getType()` of the source RA?

tischi:
> Huh, in the case of Views, can they not just delegate to the getType() of the source RA?

But I guess all of this means having to touch quite some code in imglib2, I think something that @axtimwalde was afraid of when discussing this in gitter?! Personally, as my "cloud hosted image" use-cases (and I guess more people will have such use-cases in the future) are suffering quite a bit from this issue I would still be in favor of trying to find a solution. 

@hanslovsky for the `Converters`, don't you actually have to provide a type when using them? 

For example:

```
final static public < A, B extends Type< B > > RandomAccessible< B > convert(
			final RandomAccessible< A > source,
			final Converter< ? super A, ? super B > converter,
			final B b ) // <= here
	{
		if ( TypeIdentity.class.isInstance( converter ) )
			return ( RandomAccessible< B > ) source;
		return new ConvertedRandomAccessible<>( source, converter, b );
	} 
```

And then `ConvertedRandomAccessible` has a method, which feels close to the `getType()` that we would like to have:

```
public B getDestinationType()
{
	return converted.copy();
}
```

Does that help?

hanslovsky:
@gselzer 
> Huh, in the case of Views, can they not just delegate to the getType() of the source RA?

Yes, that is exactly my point. They will have to delegate to make it any useful. That's why I don't like the `default` implementation because it will hide all those cases.

gselzer:
> Yes, that is exactly my point. They will have to delegate to make it any useful. That's why I don't like the `default` implementation because it will hide all those cases.

Well, what do you mean by 'make it any useful'? I think that your `default` RA implementation is useful, because that is what someone would do to get its type if they did not know anything more about what the implementation was. Sure, it is slow, but do you have any alternative when you have a `RA`?

gselzer:
> It will not provide any benefit over what is already there unless we implement it for all classes.

Well, it might if a given implementation has implemented it, no? If you have a `RA` of unknown implementation, you have a chance of improved performance with `getType()`, as opposed to it *always* being slow otherwise.

Anyways, I'll delegate to your judgement here, the PR is a fantastic addition either way :grin:

hanslovsky:
> Well, it might if a given implementation has implemented it, no? If you have a RA of unknown implementation, you have a chance of improved performance with getType(), as opposed to it always being slow otherwise.

Generally, I agree, as long as a simple operation like `Views.interval` or `Views.zeroMin` etc does not hide the performance improvement. This should be mitigated (to some extent) by the added implementations within imglib2 core in this PR. This will likely cover a lot of the use cases already. There may be some additional implementations, e.g. in imglib2-realtransform that should override the default implementation.

> Anyways, I'll delegate to your judgement here, the PR is a fantastic addition either way :grin:

Thank you, I appreciate it. I will leave it to the maintainers (@tpietzsch @axtimwalde @StephanPreibisch) to decide what part of the PR should go into the repo. My recommendation is to add implementations wherever possible for classes that implement `RA` or `RAI`. I will also add similar implementations for classes that implement `RRA` or `RRARI`.

hanslovsky:
I recently ran into this again, where a more efficient way to retrieve the type would have been very helpful. I rebased over current master/main and marked as ready for review.

hanslovsky:
I created this little [kscript](https://github.com/kscripting/kscript) to demonstrate the behavior + fix:
```kotlin
@file:Repository("https://maven.scijava.org/content/groups/public")
// @file:Repository("https://jitpack.io") // uncomment for getType fix
// @file:DependsOn("com.github.imglib:imglib2:feature-rai-getType-SNAPSHOT") // uncomment for getType fix
@file:DependsOn("net.imglib2:imglib2:6.2.0") // uncomment for current behavior
@file:DependsOn("net.imglib2:imglib2-cache:1.0.0-beta-17")

import net.imglib2.cache.img.CachedCellImg
import net.imglib2.cache.img.CellLoader
import net.imglib2.cache.ref.SoftRefLoaderCache
import net.imglib2.img.basictypeaccess.volatiles.array.VolatileIntArray
import net.imglib2.img.cell.Cell
import net.imglib2.img.cell.CellGrid
import net.imglib2.type.numeric.integer.IntType
import net.imglib2.util.Util

val dims = longArrayOf(10L)
val chunks = intArrayOf(1)
val grid = CellGrid(dims, chunks)

val cache = SoftRefLoaderCache<Long, Cell<VolatileIntArray>>().withLoader {
    println("Loading cell $it")
    Cell(chunks, longArrayOf(it), VolatileIntArray(1, true))
}

val img = CachedCellImg(grid, IntType(), cache, VolatileIntArray(1, true))

cache.invalidateAll(5000)
println("getTypeFromInterval")
Util.getTypeFromInterval(img)
cache.invalidateAll(5000)
println("getLinkedType")
img.createLinkedType()
println("getTypeFromInterval")
Util.getTypeFromInterval(img)
```

Output with imglib2-6.2.0:
```
getTypeFromInterval
Loading cell 0
getLinkedType
getTypeFromInterval
Loading cell 0
```

Output with this PR:
```
getTypeFromInterval
getLinkedType
getTypeFromInterval
```

src/main/java/net/imglib2/converter/readwrite/WriteConvertedRandomAccessibleInterval.java
```diff
@@ -74,4 +75,31 @@ public WriteConvertedRandomAccess< A, B > randomAccess( final Interval interval
 	{
 		return new WriteConvertedRandomAccess< A, B >( sourceInterval.randomAccess( interval ), converterSupplier );
 	}
+
+	@Override
+	public B getType() {
+		return converterSupplier.get().convert( ( Sampler< ? extends A > ) new ConstantSampler( getSource().getType() ) );
+	}
+
+	private static class ConstantSampler< T > implements Sampler< T >
```
hanslovsky, line 84: Is duplicated, could be extracted and shared between `WriteConvertedRandomAccessibleInterval` and `WriteConvertedRandomAccessible`.

src/main/java/net/imglib2/img/AbstractNativeImg.java
```diff
@@ -77,4 +78,13 @@ public T createLinkedType()
...
 	}
+
+	@Override
+	public T getType() {
+		try {
+			return linkedType.createVariable();
+		} catch ( final NullPointerException e ) {
+			return super.getType();
+		}
+	}
```
hanslovsky, line 89: This could move into the `NativeImage` interface as
```
	@Override
	public default T getType() {
		return createLinkedType().createVariable();
	}
```
Is there a performance penalty to creating a linked type vs just `createVariable()`? Is this `null` safe or should we add a `null` check ?

src/main/java/net/imglib2/util/Util.java
```diff
@@ -800,8 +806,14 @@ final static public long[] int2long( final int[] i )
...
 	 * @return - an instance of T
 	 */
+	@Deprecated
 	final public static < T, F extends Interval & RandomAccessible< T > > T getTypeFromInterval( final F rai )
 	{
+		if (rai instanceof RandomAccessibleInterval)
+			return ((RandomAccessibleInterval<T>) rai).getType();
+		// TODO can we remove generic parameter F and use
+		//      RandomAccessible<T> rai instead?
+		//      This would be a breaking change, though.
```
hanslovsky, line 816: I suggest we just leave the generic parameter as is.

src/test/java/net/imglib2/RandomAccessibleIntervalTest.java
```diff
@@ -0,0 +1,59 @@
...
+public class RandomAccessibleIntervalTest
+{
+
+	@Test
+	public void testGetType()
+	{
+		// setup
+		final RandomAccessibleInterval< ? > rai = ArrayImgs.bytes( 1 );
+		// process & test
+		assertEquals( ByteType.class, rai.getType().getClass() );
```
hanslovsky, line 57: This test is not very useful. Should we remove it?

tpietzsch:
I'm trying to push this a bit further. I added `getType()` to `IterableRealInterval` as well. Which triggered some unexpected
problems:

* IterableRealInterval and RandomAccessible cannot both have default implementations for `getType()`, otherwise we run into a bug with generic bounds: https://bugs.openjdk.org/browse/JDK-7120669
* For the same reason, the `getType()` method must be inherited from a common super-type. Possibly it is possible to avoid this by not having a default implementation at all. Something to try. On the other hand, a `Typed<T>` interface is maybe a good idea?

I also want to have `getType()` in Cursors and RandomAccesses. I'm trying to come at the problem from this side in a separate branch `sampler-getType`. In that branch, the Accessibles don't have `getType()` yet, so that we don't fall back on those out of convenience... In the end, when both branches are merged, we will see where it is better to fall back on Accessible vs to fall back on Accessors.

tpietzsch:
I think this is ready to merge.

I explored consequences and required changes in dependent (imglib2) repositories:
https://github.com/imglib/imglib2-unsafe/pull/9
https://github.com/imglib/imglib2-realtransform/pull/47
https://github.com/imglib/imglib2-cache/pull/24
https://github.com/imglib/imglib2-roi/pull/72
https://github.com/imglib/imglib2-algorithm/pull/100
https://github.com/imglib/imglib2-ij/pull/40
https://github.com/imglib/imglib2-cache-examples/pull/5
https://github.com/imglib/imglib2-imglyb/pull/11
https://github.com/imglib/imglib2-introductory-workshop/pull/4

The fact that both `IterableInterval` and `RandomAccessibleInterval` have `getType()` methods now requires API-breaking changes in imglib2-roi: https://github.com/imglib/imglib2-roi/pull/71 (but maybe having something extend `IterableInterval` and `RandomAccessibleInterval` with different types was not such a brilliant idea in the first place...)

tpietzsch:
As a follow-up, I would consider adding `T getType(T type)` to `InterpolatorFactory<T,F>`. This would allow to implement `Interpolant::getType` better. Probably the we should implement `getType()` also in `NearestNeighborSearch` (which would probably be easy for example `KDTree` provides `getType()` already).

ctrueden:
In case it's useful, here is the output of `check-branch.sh` from scijava-scripts:

0000 f27276b7629e02d352251223696527cb6d94b70d SUCCESS 15
0001 9f486c621ae6ac3e7668668d045adf1e93166ba2 SUCCESS 13
0002 dce8606a423657f93a465e11831e6f58b7901767 SUCCESS 12
0003 3c69f3ccd32e71b06c734ed1d4c91a867a008470 SUCCESS 16
0004 5b4dd4141a09b0bff28a7d73216ca76b42f8c932 SUCCESS 14
0005 42c1aec098c67a6521bd099cbe8785d8f3c0f479 FAILURE 3
0006 101d56a1acf56cb9e0a99027c182274f1af54def SUCCESS 13
0007 aa7c316881840c564356eeeb506e45f46c75e59a FAILURE 4
0008 e9893c45307ef443f9f088adc7dbfb4a3354c6a2 SUCCESS 14
0009 e28fe433133ecb7eb80b0fd1a768385e8a56edde FAILURE 4
0010 f2535358538d30967601c9aa932bc92514549016 FAILURE 3
0011 bf4bd7f89ed0df6f4d6f4a6aeba2f70c6c235035 SUCCESS 12
0012 35374c803b26548ca59e5671f895b7caa913dc07 SUCCESS 18
0013 f5d6207c4b049109a7f9c5bf6a66b4fb06486dca SUCCESS 14
0014 60f12d51ce27483917d7d59a228ae73adf3f3ce0 SUCCESS 13
0015 4a4f8e5d7b1498ba33f294ec2babdfbf9e69dbd2 SUCCESS 13
0016 f8c8d6bf0b1d864ce03c75ebf0974cf7a9793135 SUCCESS 15

It might be worth squashing the failing commits into the subsequently passing ones.

tpietzsch:
@ctrueden I rewrote the commits so that every intermediate state works now.
I wasn't aware of `check-branch.sh`. Very useful. Thanks!

src/test/java/net/imglib2/RandomAccessibleIntervalTest.java
ctrueden: I like it! A simple sanity check.

ctrueden (changes requested):
Thanks again to everyone who has been working on this! Having a `getType()` accessor will be soooo nice.

I tested this branch against SciJava Ops Image. Here are the obstacles I encountered:

There is one class, `ShuffledView`, that implements `RandomAccessibleInterval<T>` and fails to compile due to missing `getType()` implementation. (It would be nice if RAI could have a default implementation, but I see the comment about JDK-7120669, so we'll just have to live with the need to implement the new method in these situations, eh?) I implemented the method.

Next compile error:
> `java.lang.AbstractMethodError: Method net/imglib2/algorithm/neighborhood/RectangleShape$NeighborhoodsAccessible.getType()Ljava/lang/Object; is abstract`

So I checked out the matching branch for imglib2-roi and imglib2-algorithm and built those as well.

More compile-time errors:
> `.../scijava-ops-image/src/main/java/org/scijava/ops/image/image/watershed/WatershedSeeded.java:[210,28] error: incompatible types: IterableRegion<B> cannot be converted to IterableInterval<Void>`

and similar errors relating to the `IterableRegion` RAI/II split. I added the needed `.inside()` in the required spots.

Now on to runtime errors:
```java
[ERROR]   LiftFunctionsToImgTest.testLiftingArity1:64->lambda$new$0:53 » NullPointer
[ERROR]   MergeLabelingTest.testMask:113 » ClassCast class net.imglib2.roi.util.IterableRegionOnBooleanRAI cannot be cast to class net.imglib2.IterableInterval (net.imglib2.roi.util.IterableRegionOnBooleanRAI is in module net.imglib2.roi@0.15.0-SNAPSHOT of loader 'app'; net.imglib2.IterableInterval is in module net.imglib2@6.3.1-SNAPSHOT of loader 'app')
[ERROR]   ExtendRandomViewTest.extendRandomTest:75 » NullPointer
[ERROR]   RealTransformTest.regressionTest:71 » AbstractMethod Receiver class net.imglib2.realtransform.RealTransformRealRandomAccessible does not define or inherit an implementation of the resolved method 'java.lang.Object getType()' of interface net.imglib2.RealRandomAccessible.
```

Woohoo, it's our first real bug in ImgLib2! Here is an MCVE:
```java
import net.imglib2.RandomAccess;
import net.imglib2.img.Img;
import net.imglib2.img.array.ArrayImgs;
import net.imglib2.type.numeric.real.DoubleType;
import net.imglib2.view.Views;

public class Go {
  public static void main(String... args) {
    Img<DoubleType> img = ArrayImgs.doubles(10, 10);
    RandomAccess<DoubleType> il2 = Views.extendRandom(img, 0, 0).randomAccess();
  }
}
```

And the output:
```
$ javac -cp target/imglib2-6.3.1-SNAPSHOT.jar Go.java
$ java -cp target/imglib2-6.3.1-SNAPSHOT.jar:. Go
Exception in thread "main" java.lang.NullPointerException
	at net.imglib2.type.numeric.real.DoubleType.get(DoubleType.java:122)
	at net.imglib2.type.numeric.real.DoubleType.copy(DoubleType.java:181)
	at net.imglib2.type.numeric.real.DoubleType.copy(DoubleType.java:55)
	at net.imglib2.outofbounds.OutOfBoundsRandomValueFactory.create(OutOfBoundsRandomValueFactory.java:81)
	at net.imglib2.outofbounds.OutOfBoundsRandomValueFactory.create(OutOfBoundsRandomValueFactory.java:50)
	at net.imglib2.view.ExtendedRandomAccessibleInterval.randomAccess(ExtendedRandomAccessibleInterval.java:75)
	at Go.main(Go.java:10)
```

With imglib2 v6.3.0 it does not throw an exception.

So we probably should fix this before merging, and add a unit test.

Also, about versioning: I would suggest to bump the version to 7.0.0-SNAPSHOT as part of this work, because it's clearly pretty disruptive of downstream code; the major version increment serves as a signal that people should exercise more caution when upgrading.

tpietzsch:
Hmm, I ran across that problem too, for example here:
https://github.com/imglib/imglib2-algorithm/pull/100/commits/0a9dbf3c2fbf43d9cf1af2c5133a801a647ed1fe#diff-c3176d59c21f348a89eb1ee07dfec22d2cb5217d13d7b57367de8553653bc2e2L196-R200

**The problem** is that the `linkedType` returned by `NativeImg.getType()` does not have a value.
It is linked to the `Img` instance, but the `linkedType.updateContainer(...)` has never been called to link it to the `Img`s storage (e.g., the data underlying one of the Cells of a CellImg).

**The solution** is to not use `copy()` here https://github.com/imglib/imglib2/blob/5b5de3ab206278b83e2cd8749c1bd243b341e905/src/main/java/net/imglib2/outofbounds/OutOfBoundsRandomValueFactory.java#L81
but `createVariable()` instead.
(The `copy()` does `createVariable()`, followed by trying to set the value, which causes the NPE when reading the value of the type-to-be-copied).

The question is how to deal with similar problems.
One options is to require by contract that `getType()` returns a "fully realized" type. For example, `NativeImg.getType()` should then return `linkedType.createVariable()` instead of `linkedType`.

I don't like this. Actually, I was debating whether it would be better for getType to return `Class<T>` instead of a `T` instance. I landed on `T` instance, because that will cause less disruption, but also because I had the feeling that *maybe*, int the future, the instance could carry additional information. For example about the underlying access type (`Volatile` or not, backed by primitive arrays or `ByteBuffer` or something else). And my feeling is that by adding contract now, we will complicate these potential future use cases. (For no good reason, because the NPE problem is easy to fix).

Another option is to just leave it like this, let every "mis-use" blow up, and fix as they occur. I was leaning towards that, but maybe the problem is more widespread than I thought, given that we both already ran into instances.

I suggest the following:
I will try to track down all instances in imglib2 itself. Additionally, I will make the deprecated `Util.getTypeFromInterval(rai)` return `rai.getType().createVariable()` instead of `rai.getType()`. That way, old code should continue to work. Problems should occur only when the old code is updated to use `getType()` instead of the deprecated `Util.getTypeFromInterval` and we deal with them case by case.

What do you think @ctrueden? (@axtimwalde @hanslovsky)

tpietzsch:
Additionally, I will check in the `copy()` implementations for NativeTypes whether there is an associated access, and in case there is not, just return `createVariable()` without setting a value.

ctrueden (approved):
@tpietzsch I tested the updated code against scijava-ops-image, and all the tests pass now. I did have to include the updated imglib2-realtransform as well.

I then ran the pom-scijava mega-melt with the following components pinned to the respective `feature-rai-getType` branches as of today:

* imglib2
* imglib2-algorithm
* imglib2-cache
* imglib2-ij
* imglib2-realtransform
* imglib2-roi
* imglib2-unsafe

Out of 116 total components tested, 14 failed the melt:

* net.imagej:imagej-common
* net.imagej:imagej-ops
* net.imglib2:imglib2-imglyb
* net.preibisch:BigStitcher
* net.preibisch:multiview-reconstruction
* sc.fiji:bigdataviewer-core
* sc.fiji:bigdataviewer-vistools
* sc.fiji:bigwarp_fiji
* sc.fiji:labkit-pixel-classification
* sc.fiji:labkit-ui
* sc.fiji:SPIM_Registration
* sc.fiji:TrackMate
* sc.fiji:TrackMate-Skeleton
* sc.fiji:z_spacing

All errors were in one of two forms: `cannot find symbol` or `XYZ is not abstract and does not override abstract method getType() in net.imglib2.Typed`, with the missing symbols being due to the new `inside()` methods being needed to adapt certain imglib2-roi types back to `IterableInterval<Void>`.

I can post the complete build logs for these failing components, if it's useful.

tpietzsch:
> I can post the complete build logs for these failing components, if it's useful.

Yes, that would be very helpful 👍

ctrueden:
Here are the logs of the 14 failing components from my melting pot test yesterday:

[mega-melt-logs-feature-rai-getType.tar.gz](https://github.com/imglib/imglib2/files/14872699/mega-melt-logs-feature-rai-getType.tar.gz)

A `grep ERROR *.log` filters them from 7436 lines down to 302. But for some errors, it might matter whether the failure occurred when rebuilding from source, versus running the component's unit tests with the deployed binary.

Most if not all of these errors should be reproducible outside the melting pot structure by cloning the component in question, and building with `mvn clean test -Dimglib2.version=7.0.0-SNAPSHOT`, as long as you have `mvn install`ed the `feature-rai-getType` branch of imglib2 within the last 24 hours. To avoid snapshot headaches, what I usually do is change the `<version>` in the pom.xml to 999, then `mvn install` that, so that I can freely use `999` everywhere without the 24-hour refresh issue biting me, and without the enforcer plugin complaining.

tpietzsch:
@ctrueden I looked at all the errors and made draft PRs for every failing project (always branch `feature-rai-getType`)
The only problem I ran into is I think unrelated: https://github.com/juglab/labkit-ui/pull/115#issuecomment-2040561196
Apart from that there should be no more problems.

tischi:
WOW 🥳 

image


Thanks for all the work!

Issue #317, opened 2022-08-06
"Stack" RealRandomAccessibles as Composite


bogovicj:
Add a View "stacking" a list of `RealRandomAccessible<T>`s to a `RealRandomAccessible<Composite<T>>`

For example, would be useful in situations like this
https://github.com/imglib/imglib2-realtransform/issues/41

PR #318, opened 2022-08-15
"Stack" RealRandomAccessibles as Composite

bogovicj:
see #317

### DIFF
--- a/src/main/java/net/imglib2/view/Views.java
+++ b/src/main/java/net/imglib2/view/Views.java
@@ -82,6 +82,7 @@ import net.imglib2.view.composite.InflateView;
 import net.imglib2.view.composite.InterleaveView;
 import net.imglib2.view.composite.NumericComposite;
 import net.imglib2.view.composite.RealComposite;
+import net.imglib2.view.composite.RealStackCompositeView;
 
 /**
  * Create light-weight views into {@link RandomAccessible RandomAccessibles}.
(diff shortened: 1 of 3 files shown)


bogovicj:
I called this a `RealStackCompositeView`, but don't like that name

bogovicj:
I won't keep working on this until there is a more urgent use case but will leave this open in case we'd like to revisit in the future.

This needs a better name for the class, and probably would benefit from a view that does the "inverse" operation, something like:
`RealRandomAccessible<T> unstackRealComposite( RealRandomAccessible<Composite<T>> stack, int component )`
but also with a better name

bogovicj:
I'd like to revist this.

When I initially started this, I thought it might be useful for building [`DisplacementFieldTransform`s](https://github.com/imglib/imglib2-realtransform/blob/master/src/main/java/net/imglib2/realtransform/DisplacementFieldTransform.java), and I believe they still could be because somehow needs to build a `RealRandomAccessible<RealLocalizable>` to create a `DisplacementFieldTransform`.

We left this PR for a year because one (common) way to create the `RealRandomAccessible<RealLocalizable>` is by [collapsing then interpolating](https://github.com/imglib/imglib2-realtransform/blob/master/src/main/java/net/imglib2/realtransform/PositionFieldTransform.java#L186-L192). That works because some interpolators (specifically `NLinearInterpolator`s can work on `Composite`s, because `Composite`s are `NumericType`s

* [NLinearInterpolation works with `NumericType`s](https://github.com/imglib/imglib2/blob/0f83262f6a1db0b131341fe28ec1d4f495c65a7a/src/main/java/net/imglib2/interpolation/randomaccess/NLinearInterpolatorFactory.java#L49)
* [Many `Composite`s are `NumericType`s](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/view/composite/AbstractNumericComposite.java#L48)


Other interpolators will not work today, for example:

* [Lanczos](https://github.com/imglib/imglib2/blob/0f83262f6a1db0b131341fe28ec1d4f495c65a7a/src/main/java/net/imglib2/interpolation/randomaccess/LanczosInterpolatorFactory.java#L181)
* [BSpline](https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/bspline/BSplineCoefficientsInterpolator.java#L54)

because they use `RealType`s. and even [`RealComposite` types are not `RealType`s](https://github.com/imglib/imglib2/blob/0f83262f6a1db0b131341fe28ec1d4f495c65a7a/src/main/java/net/imglib2/view/composite/RealComposite.java#L54).

We might choose to make some composite types `RealType`s, but that would sacrifice some nice behavior, for example clipping. Clipping requires a min and max value, and there's no ordering of vectors.  There are at least three clipping interpolators: [Lanczos](https://github.com/imglib/imglib2/blob/0f83262f6a1db0b131341fe28ec1d4f495c65a7a/src/main/java/net/imglib2/interpolation/randomaccess/LanczosInterpolator.java#L63)  [Bspline](https://github.com/imglib/imglib2-algorithm/blob/2d8ae779fdfac1c77ccc6578d794c2f8121fc938/src/main/java/net/imglib2/algorithm/interpolation/randomaccess/BSplineInterpolator.java#L80) and [ClampingNLinearInterpolator](https://github.com/imglib/imglib2/blob/7ab23296531685b45b784c3f95719f2fe01aa945/src/main/java/net/imglib2/interpolation/randomaccess/ClampingNLinearInterpolatorFactory.java)

A straightfoward solution to interpolate each component of the vector independently, then "stack" them. Given a list of vector componets:

```java
final RealRandomAccessible<T>[] components 
```

It is possible to do this now with `BiConverter`s, but it's very verbose:

```java
final BiConverter<T, T, RealComposite<DoubleType>> conv0 = new BiConverter<T,T,RealComposite<DoubleType>>()
{
    @Override
    public void convert( T x, T y, RealComposite<DoubleType> output ) {
        output.get(0).set(x.getRealDouble());
        output.get(1).set(y.getRealDouble());
    }
};

final int N = 3;
final RealComposite<DoubleType> v = DoubleType.createVector(N);
final RealRandomAccessible<RealComposite<DoubleType>> img = Converters.convert( components[0], components[1], conv0, v);
RealRandomAccessible<RealComposite<DoubleType>> total = img;
for( int i = 2; i < N; i++ )
{
    final int j = i;
    total = Converters.convert(
            total,
            components[i],
            new BiConverter<RealComposite<DoubleType>,T,RealComposite<DoubleType>>()
            {
                @Override
                public void convert( RealComposite<DoubleType> part, T x, RealComposite<DoubleType> whole ) {

                    for( int d = 0; d < j; d++ )
                        whole.get(d).set(part.get(d));

                    whole.get(j).set(x.getRealDouble());
                }
            },
            v.copy());
}
```

This PR would 

```java
RealRandomAccessible<Composite<T>> total2 = Views.realComposite(components);
```

PR #327, opened 2022-11-21
Default implement a few methods, deprecate `copyCursor()`, and clean up.

tpietzsch:
* Default implement `Iterator.jumpFwd()`

* Default implement `RealCursor.next()`
A potential argument against this could be performance: Every call to the default `next()` goes to polymorphic `fwd()` and `get()` calls. However, I did a benchmark, and it doesn't matter. Probably because the method is so small it is inlined everywhere. Additionally, we had already implemented `next()` in `AbstractCursor`, `AbstractCursorInt`, etc, so we would have this problem already anyway.

* Default implement `IterableRealInterval.iterator()` and `IterableRealInterval.firstElement()`

* Remove implementations of `java.util.Iterator.remove()`. There is a default implementation in the `java.util.Iterator` interface now.

* Deprecate and default implement `RealCursor.copyCursor()`.
@axtimwalde already took care of `copyRandomAccess()` and `copyRealRandomAccess()` in 4ae96db3 and 054a716d, respectively.

* Remove abstract overrides of abstract methods to the exact same signature.

* Remove meaningless `final` modifiers from arguments of abstract methods.
`final` annotations of method arguments are relevant to the method implementation, but not to the caller. (It is even allowed to override methods with non-final arguments, if arguments are declared `final` in the super class.)

* Remove redundant modifiers:
In particular, remove `public` modifier from interface methods. Remove `public`, `static`, and `final` from interface fields. Remove `static` from inner enums. Remove `private` from enum constructors. These are all implicit in the respective context.

### DIFF
--- a/src/main/java/net/imglib2/RealCursor.java
+++ b/src/main/java/net/imglib2/RealCursor.java
@@ -83,7 +83,7 @@ public interface RealCursor< T > extends RealLocalizable, Sampler< T >, Iterator
 	// we must avoid doing so for now. For details, see:
 	// http://bugs.sun.com/view_bug.do?bug_id=6656332
 	// The bug is fixed in JDK7.
-	public RealCursor< T > copyCursor();
+	RealCursor< T > copyCursor();
 //@Override
 //public RealCursor< T > copy();
 }
--- a/src/main/java/net/imglib2/RealCursor.java
+++ b/src/main/java/net/imglib2/RealCursor.java
@@ -101,7 +101,7 @@ public interface RealCursor< T > extends RealLocalizable, Sampler< T >, Iterator
 	/**
 	 * Default implementation, calls {@link #fwd()} then {@link #get()}.
 	 * <p>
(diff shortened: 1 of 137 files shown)


src/main/java/net/imglib2/RealCursor.java
```diff
@@ -83,7 +86,29 @@
...
+		return copy();
+	}
+
+	@Override
+	RealCursor< T > copy();
+
+	/**
+	 * Default implementation, calls {@link #fwd()} then {@link #get()}.
+	 * <p>
+	 * Note, that {@code hasNext()} is not checked before {@codw fwd()}.
```
imagejan, line 104:
```suggestion
	 * Note, that {@code hasNext()} is not checked before {@code fwd()}.
```

tpietzsch:
@axtimwalde @StephanPreibisch I requested your review, but I think you don't have to look at the code... If you could just read the PR description and give your opinion on the general direction ...

axtimwalde:
Sounds great!

PR #331, opened 2023-06-02
Sparse matrices as Imgs

minnerbe:
I propose to include [CSR/CSC sparse arrays](https://en.wikipedia.org/wiki/Sparse_matrix#Compressed_sparse_row_(CSR,_CRS_or_Yale_format)) wrapped as Img.

## Motivation
The need for supporting traditional sparse arrays arose while working on a [spatial transcriptomics visualization library using ImgLib2](https://github.com/PreibischLab/STIM). This library needs to read and write files stored in the (Python based) [AnnData](https://anndata.readthedocs.io/en/latest/fileformat-prose.html#sparse-arrays) format, which stores certain data as [scipy CSR/CSC arrays](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_matrix.html) on disk. While an on-the-fly conversion to a dense array would be possible in most cases, it is more convenient having the sparse data represented as proper ImgLib2 Imgs.
Also, considering the popularity of CSR/CSC arrays in scientific computing in general, it's maybe of independent interest to represent and visualize them in ImgLib2. 

## Changes
I introduced the following without changing existing code:
* An abstract `SparseImg` class from which CSR and CSC style Imgs are derived. These basically wrap three 1D `ArrayImg`s (data, indices on contiguous slices of the matrix, and pointers to those slices in the data / indices array). This abstract class also features some static utility methods.
* A `SparseRandomAcess` that, given an index, finds the corresponding element in `O(log(n))`, where `n` is the size of the contiguous dimension.
* A `SparseLocalizingCursor` that traverses the whole sparse array in linear complexity and (trivially) without cache misses on the underlying `ArrayImg`s. Since any cursor needs to track the current position internally to check if the next element is zero, also `Img::cursor` returns the localizing cursor.
* A `SparseImgFactory` that creates an empty CSR/CSC array.
* A test asserting correctness of most functionality.

## Possible problems
* CSR/CSC arrays are inherently limited to 2D.
* Since adding new non-zero elements requires `O(number of current non-zero elements)` steps, this operation is usually not supported by sparse arrays. Thus, populating a new sparse matrix created by a factory is not possible, severely limiting the usefulness of `SparseImgFactory`.
* The documentation is quite sparse. ;) I can add more if requested.


minnerbe:
Since this was primarily needed to load AnnData files, an offline discussion with @tpietzsch suggested that this is too specialized to be part of core imglib.

The functionality is now implemented in a dedicated small [library for reading and writing AnnData files](https://github.com/JaneliaSciComp/n5-anndata).

PR #348, opened 2023-10-13
Draft: streamify View API

minnerbe:
As discussed with @tpietzsch, this is an attempt to create an API for `View` that's similar to Java streams. The goal would be to allow this kind of syntax:
```Java
Img<?> img = ...
Img<?> changedImg =  img.view()
		.translate(1, 2)
		.expandValue(new DoubleType(0), 1, 1)
		.permute(0, 1)
		.collect();
```

This is still work in progress. One particular limitation I encountered so far is that most methods in `Views` map `RandomAccessibleInterval` into `RandomAccessible`, i.e., they leave their domain, which makes daisy-chaining impossible.


tpietzsch:
I made a variation with multiple `...View` interfaces and wrappers in https://github.com/imglib/imglib2/tree/draft/streamify-views.

Usable like this:

https://github.com/imglib/imglib2/blob/992e3e646d53ad7d4f84cd83bf7770eb4c2164b7/src/test/java/net/imglib2/streamifiedview/StreamifiedViewsExample.java#L14-L25

minnerbe:
I added a rough draft for converting a `RaiView` to an `ArrayImg`. I probably made some errors while trying to wrestle the generic type of the view into a `NativeType`, but it seems to work for a small test example.
What do you say, @tpietzsch?

tpietzsch:
There is an update version of this PR in #364

PR #349, opened 2023-10-16
Add convenience methods for creating images from arrays

minnerbe:
The "learnathon" part of the ImgLib2/BDV hackathon in Brno showed that there is a need for an `Img`-creation API that caters to absolute beginners. The existing API for wrapping native arrays by means of the `ArrayImgs` util-class implicitly introduces the concepts of Img-backend and data-access, which seemed to keep people from exploring more awesome features.

I do think that wrapping native arrays is the best way to introduce people that are familiar with Matlab / Python / ... to ImgLib2, so I'm proposing this little wrapper API that can serve as a first exposure point for beginners:
```Java
ArrayImg<FloatType, ?> img = ArrayImgs.floats(array, dims) // previous
Img<FloatType> img = Imgs.fromArray(array, dims)           // new
```

Just to clarify again, this is meant as an educational API, so I don't see any need for deprecating the `ArrayImgs` API. This is simply inspired by the `List`-API that does a very good job of keeping implementational details hidden.

### DIFF
--- a/src/main/java/net/imglib2/img/Imgs.java
+++ b/src/main/java/net/imglib2/img/Imgs.java
@@ -1,0 +1,94 @@
+package net.imglib2.img;
+
+import net.imglib2.img.array.ArrayImgs;
+import net.imglib2.type.numeric.complex.ComplexDoubleType;
+import net.imglib2.type.numeric.complex.ComplexFloatType;
+import net.imglib2.type.numeric.integer.ByteType;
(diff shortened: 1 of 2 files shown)


minnerbe:
After discussing with @tpietzsch, it became clear that there is some potential to sustainably simplify static creation methods beyond this PR.
I created an according issue (#350) that supersedes this PR.

PR #360, opened 2024-03-09
feat: add RadialKDTreeInterpolatorFactory

bogovicj:
* and RadialKDTreeInterpolator

This PR adds an interpolator for KDTrees that takes a custom function of the (squared) radius to produce
a `RealRandomAccessible`. This is similar to but more general than the existing [`InverseDistanceWeightingInterpolatorFactory`](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/interpolation/neighborsearch/InverseDistanceWeightingInterpolatorFactory.java) which uses a particular class of functions for the radius.  Another difference - `InverseDistanceWeightingInterpolatorFactory` uses `RealType`, the classes provided by this PR can use any `NumericType`.

The demo below produces this output:
Screenshot from 2024-03-08 19-33-26 

```java
// generate points
int N = 12; // number of points
final double bigCircleRadius = 10;
final double angleDelta = 2 * Math.PI / N;
final List<RealPoint> pts = DoubleStream.iterate(0.0, x -> x + angleDelta).limit(N).mapToObj(tht -> {
	return new RealPoint(
			bigCircleRadius * Math.cos(tht),
			bigCircleRadius * Math.sin(tht));
}).collect(Collectors.toList());	

// generate values
final FloatType one = new FloatType(1f);
final ArrayList<FloatType> vals = new ArrayList<>();
IntStream.range(0, pts.size()).forEach( x -> { vals.add(one); });

// build a tree
final KDTree<FloatType> tree = new KDTree<>(vals, pts);

// a function of the squared radius for each rendered point
final double value = 2.0;
final double maxRadius = 2.0;
final double maxSqrRadius = maxRadius*maxRadius;
final DoubleUnaryOperator sqrRadiusFunction = r2 -> value * (1 + Math.cos(2 * Math.PI * r2 / maxRadius));


// create an image from the points and the radial function
final RealRandomAccessible<FloatType> image = Views.interpolate(
		tree, 
		new RadialKDTreeInterpolatorFactory<>(squaredRadiusFunction, maxRadius, new FloatType()));

// show the image
BdvStackSource<FloatType> bdv = BdvFunctions.show( 
		image, 
		Intervals.createMinSize(-20,-20, 40, 40),
		"rendered points",
		BdvOptions.options().is2D());

bdv.setDisplayRange(0, 2);
```

### DIFF
--- a/src/main/java/net/imglib2/interpolation/neighborsearch/RadialKDTreeInterpolator.java
+++ b/src/main/java/net/imglib2/interpolation/neighborsearch/RadialKDTreeInterpolator.java
@@ -1,0 +1,80 @@
+package net.imglib2.interpolation.neighborsearch;
+
+import java.util.function.DoubleUnaryOperator;
+
+import net.imglib2.KDTree;
+import net.imglib2.RealPoint;
+import net.imglib2.RealRandomAccess;
+import net.imglib2.interpolation.InterpolatorFactory;
+import net.imglib2.neighborsearch.RadiusNeighborSearch;
+import net.imglib2.neighborsearch.RadiusNeighborSearchOnKDTree;
+import net.imglib2.type.numeric.NumericType;
+
+/**
+ * A {@link RealRandomAccess} for {@link KDTree}s using a radial function used by
+ * {@link RadialKDTreeInterpolatorFactory}.
+ */
+public class RadialKDTreeInterpolator<T extends NumericType<T>> extends RealPoint implements RealRandomAccess<T> {
+
+	protected static final double minThreshold = Double.MIN_VALUE * 1000;
+
+	protected final RadiusNeighborSearch<T> search;
+	protected final double maxRadius;
+	protected final double maxSquaredRadius;
+	protected final KDTree<T> tree;
+	protected final T value;
+	protected final T tmp;
+	protected final DoubleUnaryOperator squaredRadiusFunction;
+
+	public RadialKDTreeInterpolator(
+			final KDTree<T> tree,
+			final DoubleUnaryOperator squaredRadiusFunction,
+			final double maxRadius,
+			T t) {
+
+		super(tree.numDimensions());
+
+		this.squaredRadiusFunction = squaredRadiusFunction;
+		this.tree = tree;
+		this.search = new RadiusNeighborSearchOnKDTree<T>(tree);
+		this.maxRadius = maxRadius;
+		this.maxSquaredRadius = maxRadius * maxRadius;
+		this.value = t.copy();
+		this.tmp = t.copy();
+	}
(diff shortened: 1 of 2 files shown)


bogovicj:
Another point: existing code does a nearest neighbor search but this PR uses a radius search.


 [`InverseDistanceWeightingInterpolatorFactory` implements `InterpolatorFactory< T, KNearestNeighborSearch<T>>`](https://github.com/imglib/imglib2/blob/6748f00ad03b0213b60d4ab37a25d74e94d878a8/src/main/java/net/imglib2/interpolation/neighborsearch/InverseDistanceWeightingInterpolatorFactory.java#L51C120-L51C142)
emphasis on the `KNearestNeighborSearch`.  

In this PR:
[`RadialKDTreeInterpolatorFactory` implements `InterpolatorFactory<T, KDTree<T>>`](https://github.com/bogovicj/imglib2/blob/ced765cc00a49dde2c60bcd6a00a14be516c0275/src/main/java/net/imglib2/interpolation/neighborsearch/RadialKDTreeInterpolatorFactory.java#L19) 
but internally [uses a `RadiusNeighborSearchOnKDTree`](https://github.com/bogovicj/imglib2/blob/ced765cc00a49dde2c60bcd6a00a14be516c0275/src/main/java/net/imglib2/interpolation/neighborsearch/RadialKDTreeInterpolator.java#L39) 

we could make `RadialKDTreeInterpolatorFactory` implement `InterpolatorFactory<T, RadiusNeighborSearchOnKDTree<T>>`, if that's preferable.

StephanPreibisch:
Hi, what is the math that defines what happens when the points overlap? I think there are some similarities with what I have written in STIM ... would be great to catch up on that (https://github.com/PreibischLab/STIM/blob/master/src/main/java/render/Render.java) ...

bogovicj:
@StephanPreibisch ,

Yea, the code in this PR is similar to what you have in STIM, in the same ways that it's similar to what is here in imglib2.  It's also different from what's in STIM in the ways its different from the imglib2 code. Some of your code there could do well in imglib2 core in my view.  Summary of the differences

1. general function of the squared radius
2. uses radius search instead of neighbor ([or combined radius-neighbor search](https://github.com/PreibischLab/STIM/blob/master/src/main/java/render/KNearestNeighborMaxDistanceSearchOnKDTree.java)
3. Can work with `NumericType`s (current impls use `RealType`s

The math is:

$$ \sum_{p \in P} f (r^2) v(p) $$ 

where $P$ is *all* the points within some radius. and $v(p)$ is whatever value the point $p$ has (needed to build the KDTree). and $f$ is a developer-provided function of the squared radius.

PR #371, opened 2024-10-11
fix bug in KDTreeUtils unflatten()

StephanPreibisch:
threw an ArrayIndexOutOfBoundsException for the following use-case:

// input: [1.0, 1.0, 1.0, 0.0, 1.0, 0.5, 0.0, 0.0, 0.0, 1.0], dim=2

// the should-be result with the fix
deep: [[1.0, 1.0, 1.0, 0.0, 0.0], [1.0, 0.0, 0.5, 0.0, 1.0]]

// without the fix
java.lang.ArrayIndexOutOfBoundsException: 5
	at net.imglib2.kdtree.KDTreeUtils.unflatten(KDTreeUtils.java:229)


tpietzsch:
thanks!

PR #375, opened 2025-02-12
fix: SingleDimensionPermutationTransform assertion

bogovicj:
* fixes incorrect assertion


bogovicj:
An assertion in `SingleDimensionPermutationTransform` prevents it from being applied to the first dimension. 

This PR fixes it.

PR #378, opened 2025-06-03
RealRandomAccessible: extend RandomAccessible

gselzer:
It seems to me like the `RandomAccessible` API makes perfect sense when applied to a `RealRandomAccessible`. 
* In other words, a f:R<sup>n</sup>&rarr;T **is also** a <em>f:Z<sup>n</sup>&rarr;T</em>. 

For that reason, shouldn't a `RealRandomAccessible` also be a `RandomAccessible`?


src/main/java/net/imglib2/RealRandomAccess.java
```diff
@@ -39,7 +39,7 @@
  * @author Stephan Preibisch
  * @author Stephan Saalfeld
  */
-public interface RealRandomAccess< T > extends RealLocalizable, RealPositionable, Sampler< T >
+public interface RealRandomAccess< T > extends RealLocalizable, RealPositionable, RandomAccess<T>
```
gselzer, line 42: Note that this line makes a `RealRandomAccess< T >` a `Localizable`, which it was not before. It's interesting that:
* `Localizable` extends `RealLocalizable` (i.e. if you can report your position in discrete space, then you can also report your position in real space)
* `RealPositionable` extends `Positionable` (i.e. if you can be positioned in real space, then you can also be positioned in discrete space)

Is there any historical reason for this?

ctrueden: Unfortunately yeah, because the `(Real)RandomAccess` interfaces go both ways (read and write), so neither can extend the other with violating one or the other of the two above directions.

I think that's OK, though:
* `RealRandomAccess` implements `RealLocalizable` and `RealPositionable` which means it also implements `Positionable` – i.e. you can position a `RealRandomAccess` at an integer coordinate; and
* `RandomAccess` implements `Localizable` and `Positionable` which means it also implements `RealLocalizable` – i.e. you can query a `RandomAccess` in real space if you like.

So maybe no change needed for the RA/RRA interfaces?

gselzer: > So maybe no change needed for the RA/RRA interfaces?

By RA/RRA, you mean `RandomAccessible`/`RealRandomAccessible` (i.e. not the `Access`es?)

src/main/java/net/imglib2/view/Views.java
```diff
@@ -122,11 +122,11 @@ public static < T, F extends EuclideanSpace > RealRandomAccessible< T > interpol
 	 *
 	 * @param source
 	 *            the {@link RealRandomAccessible} to be rasterized.
-	 * @return a {@link RandomAccessibleOnRealRandomAccessible} wrapping source.
+	 * @return source
 	 */
-	public static < T > RandomAccessibleOnRealRandomAccessible< T > raster( final RealRandomAccessible< T > source )
+	public static < T > RandomAccessible< T > raster( final RealRandomAccessible< T > source )
 	{
-		return new RandomAccessibleOnRealRandomAccessible<>( source );
```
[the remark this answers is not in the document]
gselzer: Emphatic yes to returning interface types!

> We could make this a separate PR though, and also break more API while we are at it.

Are you suggesting leaving the return type as is for this PR then?

gselzer:
> It wouldn't extend to `ReadRandomAccessibleRealInterval` extends `RandomAccessibleInterval` though, because intervals also work in the "read-as" direction (`Interval extends RealInterval`).

To be honest I hadn't fully thought that situation through, but yes, since `Interval` extends `RealInterval` I agree we cannot have `RRARI` extend `RAI`.

> I'm not sure about unintended side effects. For example `Views.interval(RRA)` would be possible and would "accidentally" also make the result `Iterable`. Do we want that?

I guess I don't see the problem, at least as is... `Views.interval` suggests you can get a `RAI` from a `RA`. If you pass through a `RRA` *acting* as an `RA`, you'll get back an object that is no longer in real space, but that's within the guarantees of the method...

Where it gets trickier in that case is if we later want some `Views.interval(RRA, RealInterval) -> RRARI`. In that, case if you wrote
```java
		BiConsumer<Localizable, DoubleType> c = (pos, out) -> out.setZero();
		RealRandomAccessible<DoubleType> rra = new FunctionRealRandomAccessible<>(2, c, () -> new DoubleType());
		RealRandomAccessibleRealInterval<DoubleType> rrari = Views.interval(rra, new FinalInterval(10, 10));
```
you'll get method clashing, which is unfortunate. Of course, we could just implement this interval solution on the fluent Views API side and avoid the problem...

> I'm a bit reluctant. Could you give more context on why you need this? (I know it's somehow related to imglib2-meta, but I still haven't had time to dig into that.

Yeah, my motivations with this PR come from imglib2-meta.

One of the main goals of that project is to create a simple user-facing class `Dataset`, containing a delegate `RandomAccessible` along with a metadata container, to promote metadata-aware computation. Importantly, the `Dataset` **is** a `RandomAccessible` such that it can be passed as-is to all existing SciJava/ImgLib functionality.

Ideally, the `RealDataset` type, offering the same functionality a delegate `RealRandomAccessible`, would also be a `Dataset`, but for that to work its delegate would have to be a `RandomAccessible`.

Does that provide any helpful context?

PR #380, opened 2025-07-05
Class hierarchy and interpolation of Masked types

tpietzsch:
This PR introduces the `Masked` types.

A `Masked<T>` combines a `T` value and a `double` mask.

The static methods `Masked.withConstant(...)` can be used to add constant mask to `RandomAccessible`, `RandomAccessibleInterval`, and `RealRandomAccessible`. The resulting Accesibles can be interpolated using standard nearest-neighbor interpolation, or `MaskedClampingNLinearInterpolatorFactory` which provide linear interpolation using pre-multiplied alpha masks.

Both will be used by BigDataViewer to add constant 1-masks to source `RandomAccessibleIntervals`, extend those with 0-masked values, interpolate the resulting `RandomAccessibles`, and blend them using alpha-weighted averaging.

This also works with `Volatile` types, which required making `Volatile` an interface. 

The PR also includes an unrelated (breaking) fix is to harmonize `realView().convert(...)` and `view().convert(...)` signatures. The same argument order is used for all of these now.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 
 	<groupId>net.imglib2</groupId>
 	<artifactId>imglib2</artifactId>
-	<version>7.1.6-SNAPSHOT</version>
+	<version>8.0.0-SNAPSHOT</version>
 
 	<name>ImgLib2 Core Library</name>
 	<description>A multidimensional, type-agnostic image processing library.</description>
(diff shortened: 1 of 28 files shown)


tpietzsch:
I decided to avoid the "making Volatile an interface" part for now. I will move this to a separate (draft) PR, that can later be bundled with other breaking changes that would require downstream changes anyway.
This will require a bit of additional case logic in bdv-core, but that's ok. I'll new PRs and leave this branch for reference...

PR #383, opened 2025-08-31
Updated forEachChunk example

mainSpi:
The code example was a little broken.

* `multithreaded()` had a misspelling.
* `chunk.forEach` was replaced by `chunk.forEachPixel`.
* The `chunk.forEach` line had a colon instead of a semicolon at the end.
* A more robust conversion to `IntType` was added, since converting directly from `RealType` causes a `ClassCastException`.
  * This step wasn't strictly needed, but ensuring the example works in all cases improves the dev experience.
* The lambda for `listOfSums.forEach` was missing a closing parenthesis and was replaced by a method reference, which looks nice.

### DIFF
--- a/src/main/java/net/imglib2/loops/LoopBuilder.java
+++ b/src/main/java/net/imglib2/loops/LoopBuilder.java
@@ -173,16 +173,16 @@ public class LoopBuilder< T >
 	 * <pre>
 	 * {@code
 	 *
-	 * List<IntType> listOfSums = LoopBuilder.setImages( image ).multithreaded().forEachChunk(
+	 * List<IntType> listOfSums = LoopBuilder.setImages( image ).multiThreaded().forEachChunk(
 	 *     chunk -> {
 	 *         IntType sum = new IntType();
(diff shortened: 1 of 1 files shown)


tpietzsch:
> * A more robust conversion to `IntType` was added, since converting directly from `RealType` causes a `ClassCastException`.
>   * This step wasn't strictly needed, but ensuring the example works in all cases improves the dev experience.

I would rather just put
```
RandomAccessibleInterval<IntType> image;
```
at the top of the example to make clear that it is supposed to be all `IntType`.

I think anything else distracts from the main point (illustrating usage of `forEachChunk`).
(And it still wouldn't work for *all* types, e.g., `ARGBType`).

mainSpi:
> I think anything else distracts from the main point (illustrating usage of forEachChunk).
(And it still wouldn't work for all types, e.g., ARGBType).

That sounds right. Added it.

`chunk.forEachPixel(pixel -> sum.add(pixel));` can be replaced by `chunk.forEachPixel(sum::add);` but I think it worsens readability, so I left it out.

PR #385, opened 2025-11-29
Add ScaledBrightnessARGBConverter

tinevez:
This is to enable in BDV a better display control for label images

This converter allows for better controling the visibility of ARGB segmentation. The scale value is used to alter the brightness of colors, which is better suited than altering R G and B in bulk.

This could be used in the BDV, to finely control the visibility of a segmentation overlay in the shape of an ARGB image.

### DIFF
--- a/src/main/java/net/imglib2/display/ScaledBrightnessARGBConverter.java
+++ b/src/main/java/net/imglib2/display/ScaledBrightnessARGBConverter.java
@@ -50,102 +50,11 @@ public class ScaledBrightnessARGBConverter
 		final int g = ARGBType.green( color );
 		final int b = ARGBType.blue( color );
 
-		// Convert RGB to normalized float values
-		final float rf = r / 255f;
-		final float gf = g / 255f;
-		final float bf = b / 255f;
+		final int nr = ( int ) Math.min( 255, r / scale );
+		final int ng = ( int ) Math.min( 255, g / scale );
+		final int nb = ( int ) Math.min( 255, b / scale );
 
-		final float max = Math.max( rf, Math.max( gf, bf ) );
-		final float min = Math.min( rf, Math.min( gf, bf ) );
-		final float delta = max - min;
-
-		// Calculate HSB values
-		float hue;
-		float saturation;
-		float brightness = max;
-
-		// Saturation
-		saturation = ( max == 0 ) ? 0 : delta / max;
-
-		// Hue
-		if ( delta == 0 )
-		{
-			hue = 0;
-		}
-		else if ( max == rf )
-		{
-			hue = ( ( gf - bf ) / delta ) / 6f;
-			if ( hue < 0 )
-				hue += 1f;
-		}
-		else if ( max == gf )
-		{
-			hue = ( 2f + ( bf - rf ) / delta ) / 6f;
-		}
-		else
-		{
-			hue = ( 4f + ( rf - gf ) / delta ) / 6f;
-		}
-
-		// Scale brightness
-		brightness = ( float ) Math.min( 1f, brightness / scale );
-
-		// Convert HSB back to RGB
-		return hsbToRGB( hue, saturation, brightness );
-	}
-
-	private static int hsbToRGB( final float h, final float s, final float b )
-	{
-		int r = 0, g = 0, bl = 0;
-
-		if ( s == 0 )
-		{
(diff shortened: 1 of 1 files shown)


tpietzsch:
If possible, I would like to avoid the dependency on `java.awt.Color`.

We could just inline the `Color.HSBtoRGB` and `Color.RGBtoHSB` implementations.
This would also avoid to have the (non-threadsafe) `float[] hsb` array.
But I strongly suspect that full conversion to/from HSB is not necessary to change the brightness.
I didn't look into the math (https://en.wikipedia.org/wiki/HSL_and_HSV#HSV_to_RGB etc) in detail, but from looking over it, I think there should be shortcuts. Do you want to dig into that? (Otherwise, I can also dig a bit deeper when I find time)

tpietzsch:
I couldn't resist... I think it should be more or less:

v := max(r,g,b)
m := min(1, v/scale) - v

and then

r' := min(255, r+m)
g' := min(255, g+m)
b' := min(255, b+m)

Could you try whether that works?

tinevez:
Done! 
Tested with the BDV overlaying a segmentation results as a label image. The scale properly alters the visibility of the segmentation overlay without touching the teint.

tinevez:
Here is what I have tested, with a label image overlay in the BDV, in 'Sum' mode, and adjusting the scale to 255, 122, and 0.

## The current implementation

https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/display/ScaledARGBConverter.java#L102

imglib2original1
imglib2original2
imglib2original3

## Scaling brightness via HSB conversion

https://github.com/imglib/imglib2/blob/183efc225cf3442bfe070671c1499afb2c42ed9a/src/main/java/net/imglib2/display/ScaledBrightnessARGBConverter.java#L47

HDBconverter1
HDBconverter2
HDBconverter3

## Simple RGB scaling

https://github.com/imglib/imglib2/blob/argb-converter/src/main/java/net/imglib2/display/ScaledBrightnessARGBConverter.java#L47

simpleRGBscaling1
simpleRGBscaling2
simpleRGBscaling3

## Conclusion

The simple RGB scaling gives perceptually similar results and is considerably simpler. I suggest we stick to it.

tpietzsch:
Could you try what I suggested above:
(input color: r,g,b)
v := max(r,g,b)
m := min(1, v/scale) - v
r' := min(255, r+m)
g' := min(255, g+m)
b' := min(255, b+m)
(output color: r,g,b)

I think this should be *exactly* the same as via HSB conversion.
(I just took the "convert-to-HSB-then-modify-S-then-convert-to-RGB" sequence and simplify the equations).

tinevez:
I had tried and it does not seem to work. The display is unsensitive to `scale` and the colors are wrong. Here is what I implemented

```java
	private static final int getScaledColor( final int color, final double scale )
	{
		final int r = ARGBType.red( color );
		final int g = ARGBType.green( color );
		final int b = ARGBType.blue( color );
		final int v = Math.max( r, Math.max( g, b ) );
		final double m = Math.min( 1., v / scale ) - v;

		final int nr = ( int ) Math.min( 255, r + m );
		final int ng = ( int ) Math.min( 255, g + m );
		final int nb = ( int ) Math.min( 255, b + m );

		return 0xff000000 | ( nr << 16 ) | ( ng << 8 ) | nb;
	}
```

Screenshot 2026-01-07 at 15 36 29

tpietzsch:
I looked at the HSB/RGB math again, and surprise: it comes out exactly to your "simple RGB scaling".
If I add +0.5 for rounding, e.g. `final int nr = ( int ) Math.min( 255, r /scale + 0.5 );` it also comes out numerically the same as going through HSB. Cool, I learned something :-)

However, then we can achieve the same result using the existing  `ScaledARGBConverter` and just setting min/max accordingly, that is,
`scaledARGBConverter.setMin(0);`
and
`scaledARGBConverter.setMax(255.0/scale);`

This will not slot into the BDV UI as easily, but I think anyway that this is a mis-use of the min/max sliders. (I'll comment more on the BDV PR)

imagejan:
> (I'll comment more on the BDV PR)

For context, here's the link, so others reading this don't have to repeat the search I just did 😉: https://github.com/bigdataviewer/bigdataviewer-core/pull/214

